PR #5, opened 2024-03-26
Move some scripts to framework submodule

davidhorstmann-arm:
Move `generate_test_code.py`, `test_generate_test_code.py` and `code_style.py` to the framework repository.

Synchronized with: https://github.com/Mbed-TLS/mbedtls/pull/8991


davidhorstmann-arm:
[Rebased to preserve git history of the files from the previous repo. The end result is the same]

Ryan-Everett-arm (dismissed):
Reproduced locally with the same result, LGTM!

Ryan-Everett-arm (approved):
LGTM

scripts/framework_dev/__init__.py
```diff
@@ -0,0 +1,3 @@
+# This file needs to exist to make framework_dev a package.
```
gilles-peskine-arm, line 1: I don't think `framework_dev` is right as a package name. Unless that's intended to be a sub-package of some `mbedtls` package? The idea with calling the directory `mbedtls_dev` is that you can use it in a program that isn't specific to Mbed TLS, for example a program that puts multiple crypto libraries together and uses some code from `mbedtls_dev` to obtain information about things like supported crypto mechanisms. So the name should keep `mbedtls` in it.

The current name is `mbedtls_dev` because it's for development of or related to Mbed TLS, as opposed to `mbedtls` which should be reserved for a Python package that contains wrappers around Mbed TLS's C APIs. We could keep that name here. Or maybe change it to `mbedtls_framework`?

scripts/framework_dev/c_parsing_helper.py
```diff
@@ -0,0 +1,131 @@
+"""Helper functions to parse C code in heavily constrained scenarios.
```
gilles-peskine-arm, line 1: The code in `c_parsing_helper.py`, `c_wrapper_generator.py` and `generate_psa_wrappers.py` has not been reviewed. (We rushed these files in via the restricted repository as part of a large security fix.) We need to review them, and most of the code needs to be [reviewed in the next few weeks for the client-server testing project](https://github.com/Mbed-TLS/mbedtls/issues/8961).

Moving the code to a new repository would be a good opportunity to review that code since it'll appear as new in the GitHub interface. For the review and rework to be manageable, these files should be in their own pull request.

Cc @minosgalanakis

scripts/generate_test_code.py
```diff
@@ -0,0 +1,1277 @@
+#!/usr/bin/env python3
```
gilles-peskine-arm, line 1: The first commit in the history is "Rename generate_code.py -> generate_test_code.py". It's a bit of an annoying edge case, but it would be nice to get the history from before the renaming too.

scripts/framework_dev/__init__.py
davidhorstmann-arm: I am happy with `mbedtls_dev` but I was assuming we'd want to generalise it given that it'll be used in TF-PSA-Crypto. I'll change this back to `mbedtls_dev`.

scripts/generate_test_code.py
davidhorstmann-arm: This seems to be non-trivial, but there does seem to be a [tool](https://github.com/rainlabs-eu/git-relevant-history) that may be able to do it. I'll give it a try.

ronald-cron-arm:
It should be beneficial to first move only the files in `mbedtls_dev`. I have created an issue for that:  https://github.com/Mbed-TLS/mbedtls-framework/issues/13. @davidhorstmann-arm, please have a look.

ronald-cron-arm:
Closing as superseded by more separate PRs

PR #6, opened 2024-03-27
Move some test helpers to framework repository

davidhorstmann-arm:
Move some files from `tests/src` and `tests/include/test` into the framework repo.

Synchronized with: https://github.com/Mbed-TLS/mbedtls/pull/8992


Ryan-Everett-arm (dismissed):
Reproduced locally with the same result, LGTM!

Ryan-Everett-arm (approved):
LGTM

include/alt-dummy/aes_alt.h
```diff
@@ -0,0 +1,23 @@
+/* aes_alt.h with dummy types for MBEDTLS_AES_ALT */
```
gilles-peskine-arm, line 1: `alt-dummy` is only ever going to apply to 3.6, not to 4.0 or tf-psa-crypto. So I think it should stay in 3.6. (I would normally treat it as framework material, I'm making an exception solely because it's only ever going to apply to a single consuming branch.)

davidhorstmann-arm:
Superseded by #65

Issue #7, opened 2024-04-16
Apply checker scripts to framework files


gilles-peskine-arm:
Before we move files to the framework repository, or when we first move a file of a given type to the framework repository, we should adapt checker scripts to check files in the framework repository.

* `code_style.py` for C sources
* `check_names.py` for C sources
* `check-python-files.sh` for Python sources
* more?

mpg:
Note: as the description says, this is something we should do for each first file we move of a given type / in a given directory.

This issue serves as a reminder to double-check in the end that we haven't forgotten something.

PR #8, opened 2024-04-16
Publish PSA framework simulator

gilles-peskine-arm:
Publish a PSA client-server simulator (psasim), originally written inside Arm as a proof-of-concept for the PSA Firmware Framework specification. We are currently planning to use it to test support for `MBEDTLS_PSA_CRYPTO_CLIENT`, though this is still tentative and not scheduled.

Resolves https://github.com/Mbed-TLS/mbedtls/issues/8960

Review target: reviewing the code for correctness is out of scope. We assume that the psasim code does what it claims. Reviewers must check that we are indeed taking the psasim code and the subsequent commits are correct.

Moved from https://github.com/Mbed-TLS/mbedtls/pull/8311. Unfortunately, you can't run the checker scripts (`check_files.py`, `code_style.py`, etc.) in the framework yet (https://github.com/Mbed-TLS/mbedtls-framework/issues/7), but they passed in the original.


psasim/Makefile
```diff
@@ -0,0 +1,64 @@
+CFLAGS ?=  -Wall -std=c99
+INCLUDE := -I./include/
+DESTDIR ?= /usr/local
```
valeriosetti, line 3: Do we really need to install the library system-wise? Can't we put the shared library to the `psasim` folder instead and link it from there?
Btw I tried `make install` on my Ubuntu workstation and of course it complains:
```
cp libpsaff.so /usr/local/lib/
cp: cannot create regular file '/usr/local/lib/libpsaff.so': Permission denied
```

psasim/src/service.c
```diff
@@ -0,0 +1,655 @@
...
+                }
+                mask = mask >> 1;
+            }
+        }
+
+        if ((timeout == PSA_BLOCK) && (exposed_signals > 0)) {
+            break;
+        } else {
+            /* There is no 'select' function in SysV to block on multiple queues, so busy-wait :( */
+            usleep(50000);
```
valeriosetti, line 254: My GCC compiler (`gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0`) complained that `usleep()` is implicitly declared. While looking into this warning I found that `usleep()` is actually [deprecated](https://man7.org/linux/man-pages/man3/usleep.3.html) in favor of `nanosleep()`. Shouldn't we update this?

psasim/Makefile
```diff
@@ -0,0 +1,64 @@
...
+PREFIX := libpsaff
+BUILDDIR ?= bin
```
valeriosetti, line 5: If I'm not wrong `BUILDDIR` seems not to be used anywhere (a part from the `clean` target, but that's not relevant)

psasim/tools/psa_autogen.py
```diff
@@ -0,0 +1,163 @@
...
+        policy = "const int strict_policy[32] = {"
+        qcode = "const char * psa_queues[] = { "
+        versions = "const uint32_t versions[32] = {"
+        queue_path = "/tmp/psa_service_"
+        start = False
+
+        for x in range(0, count):
+            qcode = qcode + "\"\", "
+            nsacl = nsacl + " 0,"
+            policy = policy + "0,"
```
valeriosetti, line 58:
```suggestion
            policy = policy + " 0,"
```

valeriosetti (approved):
As agreed on Slack we should ignore all the enhancements/fix comments I made so far, as those will be handled in a follow-up PS, and just focus on the `psasim` initial state. So this is fine for me :)

psasim/Makefile
tom-cosgrove-arm: As discussed, this should really just build and run in-place, like we do with the tests

PR #10, opened 2024-04-19
Psasim improvements

### LINKED ISSUE #9: Clean up `psasim`
PR https://github.com/Mbed-TLS/mbedtls-framework/pull/8 added `psasim` to this repository, but it didn't add any fix or restyle. The goal of this issue is to cleanup `psasim` to make it usable for the next steps.

valeriosetti:
This PR tries to clean-up the `psasim` to make it more usable for future MbedTLS client-server tests. 

**Note for reviewers**: I did a series of changes, some required to make the code build, but others maybe my personal preference. If you feel that I did something not correct or that there is more room for enhancements, please let me know and I will implement that.

Resolves #9.

### DIFF
--- a/psasim/test/Makefile
+++ b/psasim/test/Makefile
@@ -1,12 +1,21 @@
 INCLUDE := -I../include/ -I./psa_manifest
+LIB := -L../src -lpsaff
+
+TEST_BIN = 	psa_client \
+			psa_partition
 
 .PHONY: all clean
 
-all:
-	psa_autogen manifest.json
-	$(CC) psa_ff_bootstrap_TEST_PARTITION.c -lpsaff -o partition
-	$(CC) client.c -lpsaff -o client
+all: $(TEST_BIN)
+
+psa_client:
+	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@
+
+psa_partition: psa_ff_bootstrap_TEST_PARTITION.c
+	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
+
+psa_ff_bootstrap_%.c:
+	../tools/psa_autogen.py manifest.json
 
 clean:
-	rm -rf psa_manifest
-	rm -f client partition psa_ff_bootstrap_TEST_PARTITION.c
+	rm -f $(TEST_BIN) psa_ff_bootstrap_*.c
--- a/psasim/test/Makefile
+++ b/psasim/test/Makefile
@@ -13,10 +13,11 @@ GENERATED_H_FILES =	./psa_manifest/manifest.h \
 all: $(TEST_BIN)
 
 psa_client: $(GENERATED_H_FILES)
+psa_client: client.c
 	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@
 
 psa_partition: $(GENERATED_H_FILES)
-psa_partition: psa_ff_bootstrap_TEST_PARTITION.c
+psa_partition: psa_ff_bootstrap_TEST_PARTITION.c server.c
 	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
 
 psa_ff_bootstrap_%.c psa_manifest/%.h:
--- a/psasim/test/Makefile
+++ b/psasim/test/Makefile
@@ -19,3 +19,4 @@ psa_ff_bootstrap_%.c:
 
 clean:
 	rm -f $(TEST_BIN) psa_ff_bootstrap_*.c
+	rm -f psa_notify_* psa_service_*
--- a/psasim/test/Makefile
+++ b/psasim/test/Makefile
@@ -4,19 +4,25 @@ LIB := -L../src -lpsaff
 TEST_BIN = 	psa_client \
 			psa_partition
 
+GENERATED_H_FILES =	./psa_manifest/manifest.h \
+					./psa_manifest/pid.h \
+					./psa_manifest/sid.h
+
 .PHONY: all clean
 
 all: $(TEST_BIN)
 
-psa_client:
+psa_client: $(GENERATED_H_FILES)
 	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@
 
+psa_partition: $(GENERATED_H_FILES)
 psa_partition: psa_ff_bootstrap_TEST_PARTITION.c
 	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
 
-psa_ff_bootstrap_%.c:
+psa_ff_bootstrap_%.c psa_manifest/%.h:
 	../tools/psa_autogen.py manifest.json
 
 clean:
 	rm -f $(TEST_BIN) psa_ff_bootstrap_*.c
 	rm -f psa_notify_* psa_service_*
+	rm -f psa_manifest/*
--- a/psasim/test/Makefile
+++ b/psasim/test/Makefile
@@ -4,24 +4,24 @@ LIB := -L../src -lpsaff
 TEST_BIN = 	psa_client \
 			psa_partition
 
-GENERATED_H_FILES =	./psa_manifest/manifest.h \
-					./psa_manifest/pid.h \
-					./psa_manifest/sid.h
+GENERATED_H_FILES =	psa_manifest/manifest.h \
+					psa_manifest/pid.h \
+					psa_manifest/sid.h
+
+PARTITION_SERVER_BOOTSTRAP = psa_ff_bootstrap_TEST_PARTITION.c
 
 .PHONY: all clean
 
 all: $(TEST_BIN)
 
-psa_client: $(GENERATED_H_FILES)
-psa_client: client.c
-	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@
-
-psa_partition: $(GENERATED_H_FILES)
-psa_partition: psa_ff_bootstrap_TEST_PARTITION.c server.c
+psa_client: client.c $(GENERATED_H_FILES)
 	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
 
-psa_ff_bootstrap_%.c psa_manifest/%.h:
-	../tools/psa_autogen.py manifest.json
+psa_partition: $(PARTITION_SERVER_BOOTSTRAP) $(GENERATED_H_FILES)
+	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
+
+$(PARTITION_SERVER_BOOTSTRAP) $(GENERATED_H_FILES): manifest.json server.c
+	../tools/psa_autogen.py $<
 
 clean:
 	rm -f $(TEST_BIN) psa_ff_bootstrap_*.c
--- a/psasim/test/run_test.sh
+++ b/psasim/test/run_test.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# This is a simple bash script that tests psa_client/psa_server interaction.
+# This script is automatically executed when "make run" is launched by the
+# "psasim" root folder. The script can also be launched manually once
+# binary files are built (i.e. after "make test" is executed from the "psasim"
+# root folder).
+#
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+
 set -e
 
 function clean_run() {
--- a/psasim/test/run_test.sh
+++ b/psasim/test/run_test.sh
@@ -17,9 +17,18 @@ function clean_run() {
     ipcs | grep q | awk '{ printf " -q " $$2 }' | xargs ipcrm > /dev/null 2>&1 || true
 }
 
+# The server creates some local files when it starts up so we can wait for this
+# event as signal that the server is ready so that we can start client(s).
+function wait_for_server_startup() {
+    while [ ! -f ./psa_notify_* ]; do
+    sleep 0.1
+    done
+}
+
 clean_run
 
 ./psa_partition -k &
 SERV_PID=$!
+wait_for_server_startup
 ./psa_client
 wait $SERV_PID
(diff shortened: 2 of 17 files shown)


psasim/tools/psa_autogen.py
```diff
@@ -49,7 +49,7 @@
         policy = "const int strict_policy[32] = { "
         qcode = "const char *psa_queues[] = { "
         versions = "const uint32_t versions[32] = { "
-        queue_path = "/tmp/psa_service_"
+        queue_path = "psa_service_"
```
minosgalanakis, line 52: What would happen if we set the `TMP_FILE_BASE_PATH      "./somedir"` ? How will this python script pick it up?

psasim/tools/psa_autogen.py
```diff
@@ -146,18 +149,18 @@
...
-            bs.write(nsacl + "\n")
-            bs.write(policy + "\n")
-            bs.write(versions + "\n")
             bs.write(handlercode)
             bs.write("\n")
             bs.write("int main() {\n")
             bs.write(sigcode)
-            bs.write("   __init_psasim(psa_queues, 32, ns_allowed, versions, strict_policy);\n")
-            bs.write("   " + entry_point + "();\nfor(;;);\n}\n")
+            bs.write("    __init_psasim(psa_queues, 32, ns_allowed, versions, strict_policy);\n")
```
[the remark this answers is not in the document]
minosgalanakis: Apologies, that is correct. I though that was a log file. 

But that gives me another question, have we run the generated files againist  `code_style.py`?

minosgalanakis (approved):
Looks good overall

psasim/tools/psa_autogen.py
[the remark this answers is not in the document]
gilles-peskine-arm: `code_style.py` only runs on files that are checked into git. We explicitly exclude generated files (which are only checked in on release branches) because those tend not to respect the code style due to having lines that are too long. (Also due to historical files that have the old style of whitespace inherited from PolarSSL, but that's something we could fix easily, we just haven't taken the time to do it.)

psasim/test/server.c
```diff
@@ -47,6 +47,11 @@ int psa_sha256_main(int argc, char *argv[])
     parse_input_args(argc, argv);
     SERVER_PRINT("Starting");
 
+    int pid = fork();
+    if (pid != 0) {
+        exit(0);
+    }
```
tom-cosgrove-arm, line 53: I think we need `setsid()` in the child of fork (i.e. after line 53) as part of standard demonisation. Demonisation also usually redirects stdin, stdout and stderr to `/dev/null`, but we may not want to do that.

It might be useful to have a way to keep the server from demonising in case someone does want to run it at a terminal for debugging (this "someone" is likely to be us in a few weeks when we're trying to track down a bug somewhere!)

gilles-peskine-arm: Given that this is purely a testing tool, why would we even want daemonization? We expect the daemon's lifetime to be controlled by a test script, not by a service monitor.

valeriosetti: @tom-cosgrove-arm suggested that instead of calling the server with the final "&" to put it into the background, we can actually daemonize it (by using a specific command line argument). Based on your comment it seems that this is a bit overengineered, ins't it?

gilles-peskine-arm: When we're testing, we can often expect the server or the whole test to misbehave. So killing the server should be easy, and killing the process group should not leave a stray process. Daemonization breaks those two properties.

When we're not testing, daemonization is desirable. But when would we ever use the psasim server for purposes other than testing?

minosgalanakis: There are two discrete use cases. One is the user who is testing locally and wants the script to exit as @valeriosetti  has done, and the other is the CI which will more likely will want the server to block and hit it with clients. So I also see @gilles-peskine-arm  point.

How about a compromise?
- We restore the original functionality to block
- We add a configuration option or parameter that make it exit when client disconnects for local user testing?

If a deamon is needed we can extend those optional parameters in the future?

gilles-peskine-arm: When I mentioned testing above, I was thinking both of ad hoc runs during development, and CI scripts. For ad hoc runs, I want to run the server, perhaps in a debugger, and kill it with Ctrl+C. For CI scripts, I want to run the server, and `kill(1)` or `kill(2)` it when done.

> We add a configuration option or parameter that make it exit when client disconnects for local user testing?

Hmmm, maybe. We'll have some tests where we want to exercise persistence between client connections, but most tests won't do that. On the other hand, when running a test suite, I think we'll want to run the server once for the whole suite (to avoid a relatively large overhead per test case, and to avoid having to code up something to make the client autostart the server), but have the client connect separately in each test case (because each test case calls `psa_crypto_init` and `mbedtls_psa_crypto_free`). So I don't see a lot of use for an auto-exit on disconnect.

valeriosetti: Ok, so if I understood correctly I just need to remove the last commit (going back to https://github.com/Mbed-TLS/mbedtls-framework/pull/10/commits/e91126b960e6b7e4c485d1a83266a58cc986ea88) and there I just need to invert the logic meaning of "k" parameter. Now the default behavior will be keep server alive if no "k" param is defined, otherwise "kill on client disconnection".

psasim/tools/psa_autogen.py
valeriosetti: I kept only the filename in the python script (and in the generated C file, of course). The full path is created in `__init_psasim()` in `framework/psasim/src/service.c` by prepending `TMP_FILE_BASE_PATH`. The worst thing that can happen IMO is that the specified folder does not exists. In  this case we should adjust Makefiles to fill this gap, but I don't know if it's worth it. Wdyt?

psasim/test/run_test.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
minosgalanakis, line 2: Can we add a header on this file?

    # Copyright The Mbed TLS Contributors
    # SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later

And or even a help line to assist people running it for the first time?

minosgalanakis (approved):
LGTM

psasim/test/Makefile
```diff
@@ -1,12 +1,22 @@
...
 .PHONY: all clean
 
-all:
-	psa_autogen manifest.json
-	$(CC) psa_ff_bootstrap_TEST_PARTITION.c -lpsaff -o partition
-	$(CC) client.c -lpsaff -o client
+all: $(TEST_BIN)
+
+psa_client:
+	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@
```
tom-cosgrove-arm, line 12: This is missing a dependency when run the first time

```
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C test CFLAGS="-Wall -Werror -std=c99 -D_XOPEN_SOURCE=1 -D_POSIX_C_SOURCE=200809L"
cc -I../include/ -I./psa_manifest -Wall -Werror -std=c99 -D_XOPEN_SOURCE=1 -D_POSIX_C_SOURCE=200809L client.c -L../src -lpsaff -o psa_client
client.c:10:10: fatal error: 'psa_manifest/sid.h' file not found
#include "psa_manifest/sid.h"
         ^~~~~~~~~~~~~~~~~~~~
```

So something like this is needed

```
diff --git a/psasim/test/Makefile b/psasim/test/Makefile
index b5044b5..4b9d023 100644
--- a/psasim/test/Makefile
+++ b/psasim/test/Makefile
@@ -8,13 +8,15 @@ TEST_BIN = 	psa_client \
 
 all: $(TEST_BIN)
 
+psa_client: ../psa_manifest/sid.h
+
 psa_client:
 	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@
 
 psa_partition: psa_ff_bootstrap_TEST_PARTITION.c
 	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
 
-psa_ff_bootstrap_%.c:
+../psa_manifest/sid.h psa_ff_bootstrap_%.c:
 	../tools/psa_autogen.py manifest.json
 
 clean:
```

psasim/test/run_test.sh
```diff
@@ -0,0 +1,25 @@
...
+function clean_run() {
+    pkill psa_partition || true
+    pkill psa_client || true
+    ipcs | grep q | awk '{ printf " -q " $$2 }' | xargs ipcrm > /dev/null 2>&1 || true
+}
+
+clean_run
+
+./psa_partition &
```
tom-cosgrove-arm, line 22: This test should run the server in such a way that it terminates when the client disconnects, to leave this script self-contained (i.e. not leave the server running after the client, with the script waiting)

valeriosetti: Sorry, I changed the behavior of the "k" parameter recently, but then I forgot to update this launch command.

psasim/test/Makefile
valeriosetti: Oh, I launched "make run" several times now, but this never happened to me. To me this looks like a make order, isn't it? Just out of my curiosity: do you have any idea of why this never happened to me?

psasim/test/Makefile
```diff
@@ -4,19 +4,25 @@ LIB := -L../src -lpsaff
...
+GENERATED_H_FILES =	./psa_manifest/manifest.h \
+					./psa_manifest/pid.h \
+					./psa_manifest/sid.h
+
 .PHONY: all clean
 
 all: $(TEST_BIN)
 
-psa_client:
+psa_client: $(GENERATED_H_FILES)
```
tom-cosgrove-arm, line 15: If we're going to add explicit dependencies when there's the compilation rules, can we include `client.c` in this list please :)

psasim/test/Makefile
tom-cosgrove-arm: I assume at some point you got `sid.h` created; my tree didn't have this until I added the dependency

psasim/test/Makefile
valeriosetti: I added those explicit dependencies because those need to be auto-generated. I didn't add already existing source files because I assumed that this was not necessary.
Following you suggestion then I would also need to check for `server.c` when building `psa_partition`, isn't it?

psasim/test/Makefile
```diff
@@ -13,10 +13,11 @@ GENERATED_H_FILES =	./psa_manifest/manifest.h \
 all: $(TEST_BIN)
 
 psa_client: $(GENERATED_H_FILES)
+psa_client: client.c
 	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@
 
 psa_partition: $(GENERATED_H_FILES)
-psa_partition: psa_ff_bootstrap_TEST_PARTITION.c
+psa_partition: psa_ff_bootstrap_TEST_PARTITION.c server.c
```
tom-cosgrove-arm, line 20: Actually, this isn't quite right - if `server.c` changes then we should re-run `psa_autogen.py` (since that script reads all the files in the directory, and gets the entry point from it).

`psa_autogen.py` then re-writes `psa_ff_bootstrap_TEST_PARTITION.c`, and we then want to compile it.

So the `server.c` dependency should really be for the rule that runs `psa_autogen.py` (i.e. at (new) line 23)

psasim/test/Makefile
```diff
@@ -13,10 +13,11 @@ GENERATED_H_FILES =	./psa_manifest/manifest.h \
...
 	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
 
 psa_ff_bootstrap_%.c psa_manifest/%.h:
```
tom-cosgrove-arm, line 23: Why not use `GENERATED_H_FILES` here? (and: this is where `server.c` should be listed, I think)

psasim/test/Makefile
```diff
@@ -20,7 +20,10 @@ psa_partition: $(GENERATED_H_FILES)
 psa_partition: psa_ff_bootstrap_TEST_PARTITION.c
 	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@
 
-psa_ff_bootstrap_%.c psa_manifest/%.h: server.c
+psa_ff_bootstrap_%.c: server.c
```
tom-cosgrove-arm, line 23: The `%` here is really being used as a wildcard/shortcut cheat, rather than as part of an pattern rule (where there would be something like `%.o : %.c`). In that case, and since we've already got an instance of the `psa_ff_bootstrap_%.c` pattern written out a few lines above, maybe we should define a variable for `psa_ff_bootstrap_TEST_PARTITION.c` and use it together with `$(GENERATED_H_FILES)` to not repeat the `psa_autogen.py` recipe.

What do you think? (And sorry this is taking longer than expected)

tom-cosgrove-arm: So we've then got something like

```
PARTITION_SERVER_BOOTSTRAP = psa_ff_bootstrap_TEST_PARTITION.c

psa_client: client.c $(GENERATED_H_FILES)
	$(CC) $(INCLUDE) $(CFLAGS) client.c $(LIB) -o $@

psa_partition: $(PARTITION_SERVER_BOOTSTRAP) $(GENERATED_H_FILES)
	$(CC) $(INCLUDE) $(CFLAGS) $< $(LIB) -o $@

$(PARTITION_SERVER_BOOTSTRAP) $(GENERATED_H_FILES): server.c
	../tools/psa_autogen.py manifest.json
```

valeriosetti: > And sorry this is taking longer than expected

No worries at all!! Reviews are meant to improve code and this is exactly what it is happening here :)
Thanks for the tips

tom-cosgrove-arm (approved):
LGTM - thanks

minosgalanakis (approved):
LGTM

minosgalanakis:
We should wait for [#9088](https://github.com/Mbed-TLS/mbedtls/pull/9088) to be merged before we merge this PR.

tom-cosgrove-arm:
Note: this code will be merged into the main Mbed TLS repository

tom-cosgrove-arm:
Replaced by https://github.com/Mbed-TLS/mbedtls/pull/9097 in the main repo

PR #20, opened 2024-05-28
Script to search the outcome file for configurations with given requirements

gilles-peskine-arm:
Resolves https://github.com/Mbed-TLS/mbedtls/issues/9195

To try it out, grab an outcome file from a nightly.

Testing: I ran `pylint` and `mypy` manually. There is no corresponding PR in the main repo because nothing in the main repo depends on it, this is an entry-point script intended for use by maintainers.

### DIFF
--- a/scripts/search_outcomes_config.py
+++ b/scripts/search_outcomes_config.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python
+"""Search an outcome file for configurations with given settings.
+
+Read an outcome file and report the configurations in which test_suite_config
+runs with the required settings (compilation option enabled or disabled).
+"""
+
+import argparse
+import os
+import re
+import subprocess
+from typing import Dict, FrozenSet, Iterator, List, Set
+
+
+def make_regexp_for_settings(settings: List[str]) -> str:
+    """Construct a regexp matching the interesting outcome lines.
+
+    Interesting outcome lines are from test_suite_config where the given
+    setting is passing.
+
+    We assume that the elements of settings don't contain regexp special
+    characters.
+    """
+    return (r';test_suite_config[^;]*;Config: (' +
+            '|'.join(settings) +
+            r');PASS;')
+
+def run_grep(regexp: str, outcome_file: str) -> List[str]:
+    """Run grep on the outcome file and return the matching lines."""
+    env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


scripts/search_outcomes_config.py
```diff
@@ -0,0 +1,222 @@
...
+    """Extract the configuration data from outcome lines.
+
+    The result maps a configuration name to the list of passing settings
+    in that configuration.
+    """
+    config_data = {} #type: Dict[str, Set[str]]
+    for line in outcome_lines:
+        m = OUTCOME_LINE_RE.match(line)
+        # make_regexp_for_settings() arranges to only return lines that
+        # should match here.
```
yanesca, line 61: Why do we need this comment here? I am not sure how it supposed to be helpful in this context. I mean when reading this code, where might be the question raised to which this is the answer?

yanesca (dismissed):
I tested the script locally and it works correctly. The source looks good to me.

scripts/search_outcomes_config.py
gilles-peskine-arm: This comment explains why the assertion on the line below is always expected to succeed.

yanesca: Oh, I didn't realise that it was trying to explain that.

It seems to me that`make_regexp_for_settings` requires a single `;` character right before `test_suite_config`. `OUTCOME_LINE_RE` requires another `;` in the preceding string. If the string preceding `;test_suite_config` doesn't have a `;` it will match with `make_regexp_for_settings` but won't match with `OUTCOME_LINE_RE` and the assertion will fail. What am I missing here?

gilles-peskine-arm: We'll have an assert failure if the file doesn't have the expected format of an outcome file. I think that's reasonable for a developer-facing script.

yanesca (approved):
LGTM

bjwtaylor (approved):

gilles-peskine-arm:
I labelled this pull request "needs-preceding-pr" but there is no indication of what it needs and looking back I can't see what it would need. It's just a new script that doesn't consume anything else except for one long preexisting Python module. It has passed the CI. So I'm going ahead and removing the label "needs-preceding-pr". This makes the PR ready to merge, so I'm merging it.

PR #22, opened 2024-06-04
Fix generation of PSA test wrappers for PAKE

ronald-cron-arm:
Associated to https://github.com/Mbed-TLS/mbedtls/pull/9185

PR #25, opened 2024-06-07
Feature: code wrappers subpackage

minosgalanakis:
This PR adds the placeholder directroy for the code-generation wrappers, refactors `generate_psa_wrappers.py` and add some minor utility functions in `c_parsing_helper.py`

In case it gets hidden by GitHub, the PR to incorporate this to the main repo is https://github.com/Mbed-TLS/mbedtls/pull/9224


scripts/mbedtls_framework/code_wrapper/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+_p = ["tests", "scripts", "code_wrapper"]
+
+for _ in range(5):
+    try:
+        sys.path.append(os.path.join(*_p))
+        import psa_test_wrapper, psa_buffer, psa_wrapper
+    except ModuleNotFoundError:
+        _p = ["../"] + _p
+        continue
```
tom-cosgrove-arm, line 18: Missing newline at end of file

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -129,3 +151,16 @@ def read_function_declarations(functions: Dict[str, FunctionInfo],
                                        return_type,
                                        name,
                                        arguments)
+_C_TYPEDEF_DECLARATION_RE = r'typedef (?:struct ){0,1}(?P<type>\w+) (?P<name>\w+)'
```
tom-cosgrove-arm, line 154: Could we have a blank line to separate this code from the function above? And maybe get rid of some of the many blank lines at line 157?

scripts/mbedtls_framework/code_wrapper/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+
```
valeriosetti, line 19: I don't know why, but on my PC I have 2 empty lines at the end of this file, whereas here on github I see only 1. In case there are really 2 lines I would ask to remove 1 whereas if it's already correct, please ignore this comment.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -60,18 +62,37 @@ def __init__(self, #pylint: disable=too-many-arguments
...
         self.name = name
         self.arguments = [ArgumentInfo(arg) for arg in arguments]
+        self.doc = doc
 
     def returns_void(self) -> bool:
         """Whether the function returns void."""
         return bool(self.VOID_RE.search(self.return_type))
 
+    def __str__(self) -> str:
+        _str_args = [str(a) for a in self.arguments]
```
gilles-peskine-arm, line 81: Minor: this shouldn't be an underscore-prefixed variable name by any of the conventions I know (class members that child classes should not access, and potentially unused variables).

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -60,18 +62,37 @@ def __init__(self, #pylint: disable=too-many-arguments
...
+        _str = "{} {} {}({})".format(" ".join(self.qualifiers),
```
gilles-peskine-arm, line 82: Minor: This should be called `str_`, per [PEP 8](https://peps.python.org/pep-0008/#descriptive-naming-styles), if the intent is to call it `str` without shadowing the built-in function by that name.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -60,18 +62,37 @@ def __init__(self, #pylint: disable=too-many-arguments
...
+                                     self.return_type, self.name,
+                                     ", ".join(_str_args)).strip()
+        _str = FunctionInfo._c_align__(_str)
+        return self.doc + "\n" + _str
+
+    @staticmethod
+    def _c_align__(in_str: str) -> str:
```
gilles-peskine-arm, line 89: Please document this method. And rename it? It doesn't actually seem to do any aligning.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -129,3 +151,16 @@ def read_function_declarations(functions: Dict[str, FunctionInfo],
                                        return_type,
                                        name,
                                        arguments)
+
+_C_TYPEDEF_DECLARATION_RE = r'typedef (?:struct ){0,1}(?P<type>\w+) (?P<name>\w+)'
+
+def read_typedefs(filename: str)->  Dict[str, str]:
+    """ Extract type definitions in a {typedef name: primitiv type} dictionary"""
```
gilles-peskine-arm, line 158: Typo: primitiv → primitive. But that's not actually a primitive type, it's just whatever the type was defined as an alias of. Only for certain typedefs, given the regex. Can you please document which typedefs are caught here?

scripts/generate_psa_wrappers.py
```diff
@@ -10,225 +10,14 @@
 ### for robustness and maintainability.
 
 import argparse
-import itertools
```
gilles-peskine-arm, line 13: The new file containing `class BufferParameter` and so on needs to be in the same repository for two reasons: because it'll apply to all branches/repos so it needs to be in the framework repository, and also so that `git diff --color-moved` shows it as a move, not as deletion. I've stopped reviewing at a8d6b102a2b6dbb3e2a71f5dcebc51bda5af6632 because as it stands now, it's just appearing to delete code that needs to stay.

gilles-peskine-arm (changes requested):
Unfinished review because `psa_buffer.py` needs to be in `mbedtls_framework`.

scripts/mbedtls_framework/c_parsing_helper.py
minosgalanakis: I actually want to override the built-in function. It serves no use as is, but its helpfull for print debugging or for writing onto a c file if we override the `__str__` representation.

gilles-peskine-arm: One of us is not understanding the other, I'm not sure which.

I know that the `__str__` method customizes the behavior of `str` on objects of this class. But it's irrelevant to my comment.

My comment is about the local variable `_str`. The widely adopted conventions recorded in PEP 8 state that local variables should not begin with an underscore, except to indicate that they are possibly uninitialized (which stops linters such as pylint warning about that). So the variable should not be called `_str`.

It appears that a natural name for this variable would be `str`. However, `str` is not a good variable name for a different reason: this is a built-in function, and while it's possible to override built-in functions, it can have very surprising effects if you accidentally both override a built-in function and attempt to use it. Hence PEP 8 recommends to not override the names of built-in functions, and if you want that name, to **append** (not prepend) an underscore. Thus the variable should be called `str_` (or something else altogether like `string`).

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -60,18 +62,39 @@ def __init__(self, #pylint: disable=too-many-arguments
...
     def returns_void(self) -> bool:
         """Whether the function returns void."""
         return bool(self.VOID_RE.search(self.return_type))
 
+    def __str__(self) -> str:
+        str_args = [str(a) for a in self.arguments]
+        str_text = "{} {} {}({})".format(" ".join(self.qualifiers),
+                                         self.return_type, self.name,
+                                         ", ".join(str_args)).strip()
+        str_text = FunctionInfo._c_wrap__(str_text)
```
gilles-peskine-arm, line 85: Why not `self._c_wrap__`?

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -60,18 +62,39 @@ def __init__(self, #pylint: disable=too-many-arguments
...
+        return self.doc + "\n" + str_text
+
+    @staticmethod
+    def _c_wrap__(in_str: str, line_len: int = 80) -> str:
+        """ Will wrap function  strings over line_len length and
+            ident them in alignment to the opening parenthesis """
```
gilles-peskine-arm, line 91: Minor: PEP 8 and our own habits say the first paragraph should be a single line. Also, no spaces around it and a full stop at the end of the sentence.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -60,18 +62,39 @@ def __init__(self, #pylint: disable=too-many-arguments
...
+        str_args = [str(a) for a in self.arguments]
+        str_text = "{} {} {}({})".format(" ".join(self.qualifiers),
+                                         self.return_type, self.name,
+                                         ", ".join(str_args)).strip()
+        str_text = FunctionInfo._c_wrap__(str_text)
+        return self.doc + "\n" + str_text
+
+    @staticmethod
+    def _c_wrap__(in_str: str, line_len: int = 80) -> str:
+        """ Will wrap function  strings over line_len length and
```
gilles-peskine-arm, line 90: What's a “function string”?

gilles-peskine-arm, line 90: Going by the context, this function is used to format function prototypes. Why not keep it simple and put one argument per line, like we're doing in the C sources? That avoids varying indentation based on the function name and zigzagging to read parameters in arbitrary groups.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -129,3 +153,17 @@ def read_function_declarations(functions: Dict[str, FunctionInfo],
                                        return_type,
                                        name,
                                        arguments)
+
+_C_TYPEDEF_DECLARATION_RE = r'typedef (?:struct ){0,1}(?P<type>\w+) (?P<name>\w+)'
+
+def read_typedefs(filename: str)->  Dict[str, str]:
```
gilles-peskine-arm, line 159: This only captures some typedefs. Why those? What will happen if we edit a header and the typedef no longer follows that format, and how much of a risk is there that this happens?

I don't know what this function is meant to be used for, and currently it isn't used in #25 or Mbed-TLS/mbedtls#9224.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -129,3 +153,17 @@ def read_function_declarations(functions: Dict[str, FunctionInfo],
                                        return_type,
                                        name,
                                        arguments)
+
+_C_TYPEDEF_DECLARATION_RE = r'typedef (?:struct ){0,1}(?P<type>\w+) (?P<name>\w+)'
```
gilles-peskine-arm, line 157: Why isn't the `struct` captured in `type`?

gilles-peskine-arm, line 157: KISS:
```suggestion
_C_TYPEDEF_DECLARATION_RE = r'typedef (?:struct )?(?P<type>\w+) (?P<name>\w+)'
```

gilles-peskine-arm (changes requested):
I've reviewed the changes in `scripts/mbedtls_framework/c_parsing_helper.py`.

For the rest, there's a big bunch of code that disappears in a8d6b102a2b6dbb3e2a71f5dcebc51bda5af6632 and reappears with modifications in 4d3153b5589f5ab4ab3f459a690502e279046301. This makes it effectively necessary to review the whole code from scratch, which is a waste of time.

Please restructure the commits so that code that is moved appears as a move with `git diff --color-moved SHA`, and changes to the content are done in separate commits. For example, in https://github.com/Mbed-TLS/mbedtls/pull/8930, “docs: Move TLS 1.3 early data doc to a dedicated file” is a separate commit b372b2e5bb2949a9e96f7f152a69f9db63862894 with no content changes.

scripts/mbedtls_framework/c_parsing_helper.py
minosgalanakis: Reading simple typedefs is something that could be usefull for a `FunctionInfo` parsing class. 

It is currently not being used, but will be used when parsing psa headers for generating serialisation logic. It can also be very usefull for debugging purposes since it generates a simple to read dictionary e.g

    {'mbedtls_svc_key_id_t': 'psa_key_id_t',
     'psa_algorithm_t': 'uint32_t',
     'psa_dh_family_t': 'uint8_t',
     'psa_ecc_family_t': 'uint8_t',
     'psa_key_attributes_t': 'psa_key_attributes_s',
     'psa_key_derivation_step_t': 'uint16_t',
     'psa_key_id_t': 'uint32_t',
     'psa_key_lifetime_t': 'uint32_t',
     'psa_key_location_t': 'uint32_t',
     'psa_key_persistence_t': 'uint8_t',
     'psa_key_production_parameters_t': 'psa_key_production_parameters_s',
     'psa_key_slot_number_t': 'uint64_t',
     'psa_key_type_t': 'uint16_t',
     'psa_key_usage_t': 'uint32_t',
     'psa_status_t': 'int32_t'}

There is very little risk, since we don't change psa headers that much, and evem when we do, the data structures are very rarely updated.  Also this will be used in test code generation so whoever uses it needs to make sure it is capturing everything and update it accordingly. 

For now the simple parsing should work.

scripts/mbedtls_framework/c_parsing_helper.py
minosgalanakis: Because we do not have to capture it to know it is a structure.

scripts/mbedtls_framework/c_parsing_helper.py
minosgalanakis: Because there is no usage of self in the method which will fail pylint. Plus string formatting methods are usefull to be static so others can use it directly without instantation.

scripts/mbedtls_framework/c_parsing_helper.py
minosgalanakis: This is placing an argument per line as we do in the source code ?

Namely it replaces normal commas with a `" " * pad_no + ",\n" `in the `in_str = in_str.replace(",", padded_comma)` method.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,236 @@
...
+        for header_name in headers:
+            header_path = self.rel_path(header_name)
+            c_parsing_helper.read_function_declarations(self.functions, header_path)
+
+    def rel_path(self, filename: str, path_list: List[str] = ['include', 'psa']) -> str:
+        """ Return the estimated path in relationship to the.
+            mbedtls_root. The method allows overriding the targetted sub-directory.
+            Currently the default is set to mbedtls_root/include/psa """
+        
+        # Temporary, while Mbed TLS does not just rely on the TF-PSA-Crypto
```
minosgalanakis, line 99: @ronald-cron-arm Please note that this is a refactor of the patch you introduced in [PR26]( https://github.com/Mbed-TLS/mbedtls-framework/pull/26/commits/030b14c2bce1dff5bd28b08b2c00b6bc1fdd66d5)

This should not break https://github.com/Mbed-TLS/mbedtls/pull/9247

valeriosetti (dismissed):

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,252 @@
...
+    "not_implemented": frozenset([
+        'psa_crypto_driver_pake_inputs_t *',
+        'psa_pake_cipher_suite_t *',
+    ]),
+    "function_guards": {
+        'mbedtls_psa_register_se_key': 'defined(MBEDTLS_PSA_CRYPTO_SE_C)',
+        'mbedtls_psa_inject_entropy': 'defined(MBEDTLS_PSA_INJECT_ENTROPY)',
+        'mbedtls_psa_external_get_random': 'defined(MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG)',
+        'mbedtls_psa_platform_get_builtin_key': 'defined(MBEDTLS_PSA_CRYPTO_BUILTIN_KEYS)',
+        'psa_crypto_driver_pake_get_cipher_suite' : 'defined(PSA_WANT_ALG_SOME_PAKE)',
```
minosgalanakis, line 42: @ronald-cron-arm There are the changes introduced by [PR22](https://github.com/Mbed-TLS/mbedtls-framework/pull/22)

ronald-cron-arm: Yes this looks similar to what has been done in PR 22.

valeriosetti (dismissed):

scripts/mbedtls_framework/code_wrapper/psa_test_wrapper.py
```diff
@@ -31,7 +31,7 @@ class PSATestWrapper(PSAWrapper):
     _WRAPPER_NAME_PREFIX = 'mbedtls_test_wrap_'
     _WRAPPER_NAME_SUFFIX = ''
 
-    _PSA_WRAPPER_INCLUDES = ['<psa/crypto.h>',
+    _PSA_WRAPPER_INCLUDES = ['<psa/crypto.h>\n',
```
minosgalanakis, line 34: This newline is not required but alligns the generated files to what is already comitted

    #include <psa/crypto.h>

    #include <test/memory.h>
    #include <test/psa_crypto_helpers.h>
    #include <test/psa_test_wrappers.h>

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+    _PSA_WRAPPER_INCLUDES = ['<psa/crypto.h>']
+
+    def __init__(self,
+                 output_h_f: str,
+                 output_c_f: str,
+                 in_headers:  Collection[str] = DEFAULTS["input_headers"],
+                 config: Dict[str, Any]= DEFAULTS) -> None:
+
+        super().__init__()
+        self._FUNCTION_GUARDS = super()._FUNCTION_GUARDS.copy()
```
gilles-peskine-arm, line 76: We shouldn't be overriding a class-level constant in an instance. That will go horribly wrong if the superclass has a class method that goes `cls._FUNCTION_GUARDS`, for example.

When I wrote the code originally, my idea was that the class would know the structure of the wrapper code, and the instance would control things like the list of wrapped functions, the output destination, and some elements injected at fixed positions. If this needs to change because some scenarios require changing the structure of the wrapper code, that's fine, but we should revise all the classes accordingly. However, it's not clear to me why this is changing here. What requires this extra flexibility?

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+        self.in_headers = in_headers
+        self.out_c_f = output_c_f
```
gilles-peskine-arm, line 78: Can we please have consistent names? Either `out_c_f` or `output_c_f`.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+        self.out_h_f = output_h_f
+
+        self.mbedtls_root = build_tree.guess_mbedtls_root()
+        self.read_config(config)
+
+        if in_headers:
+            self.read_headers(in_headers)
```
gilles-peskine-arm, line 85: An empty `in_headers` isn't a special case here.

```suggestion
        self.read_headers(in_headers)
```

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+    _WRAPPER_NAME_PREFIX = 'mbedtls_test_wrap_'
+    _WRAPPER_NAME_SUFFIX = ''
+
+    _PSA_WRAPPER_INCLUDES = ['<psa/crypto.h>']
+
+    def __init__(self,
+                 output_h_f: str,
+                 output_c_f: str,
+                 in_headers:  Collection[str] = DEFAULTS["input_headers"],
+                 config: Dict[str, Any]= DEFAULTS) -> None:
```
gilles-peskine-arm, line 73: `Any` is a code smell. See my comment on `DEFAULTS` in `psa_wrapper.py`.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+        self.out_h_f = output_h_f
+
+        self.mbedtls_root = build_tree.guess_mbedtls_root()
+        self.read_config(config)
+
+        if in_headers:
+            self.read_headers(in_headers)
+
+    def read_config(self, cfg: Dict[str, Any])-> None:
+        """Configure instance's parameters based on a module specific default config """
```
gilles-peskine-arm, line 88: I don't understand “module specific default config”. What does “default” refer to? Settings in `cfg` override the current settings in `self`, so this method is potentially overriding defaults, not applying defaults.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+
```
gilles-peskine-arm, line 89: Why isn't `cfg['input_headers']` also applied here?

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+        self._CPP_GUARDS = PSAWrapper.parse_def_guards(cfg["define_guards"])
+        self._SKIP_FUNCTIONS = cfg["skip_list"]
+        self._FUNCTION_GUARDS.update(cfg["function_guards"]) # type: ignore[arg-type]
+        self._NOT_IMPLEMENTED = cfg["not_implemented"]
+
+    def read_headers(self, headers: Collection[str]) -> None:
+        """ Reads functions from source header files into a Dict[str, FunctionInfo] """
```
gilles-peskine-arm, line 96: There's no `Dict[str, FunctionInfo]` in the interface of this method. ITYM something like
```suggestion
        """ Reads functions to be wrapped from source header files."""
```

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+    # Utility Methods
+    @staticmethod
+    def parse_def_guards(def_list: Collection[str])-> str:
+        """ Parse an input list of format ["HASH_DEFINE", "!HASH_DEFINE2" ] and generate a
+            c compatible defined(HASH_DEFINE) && !defined(HASH_DEFINE) syntax string"""
+
+        output = ""
+        _dl = ["defined({})".format(n) if n[0] != "!" \
+                     else "!defined({})".format(n[1:]) for n in def_list]
+        # Split the list in chunks of 2 and add new lines
```
gilles-peskine-arm, line 125: It would be simpler both in the code and for readers to put one name per line.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+        if os.path.isdir(os.path.join(self.mbedtls_root, 'tf-psa-crypto')):
+            path_list = ['tf-psa-crypto' ] + path_list
+            return os.path.join(self.mbedtls_root, *path_list, filename)
+
+        return os.path.join(self.mbedtls_root, *path_list, filename)
+
+    # Utility Methods
+    @staticmethod
+    def parse_def_guards(def_list: Collection[str])-> str:
+        """ Parse an input list of format ["HASH_DEFINE", "!HASH_DEFINE2" ] and generate a
```
gilles-peskine-arm, line 119: Please don't use the uncommon nickname “hash define”, especially in code where “hash” normally refers to a cryptographic hash algorithm. The official and common name is “[[C] preprocessor] macro”. We sometimes use the somewhat common nickname “[[C] preprocessor] symbol”.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+            c compatible defined(HASH_DEFINE) && !defined(HASH_DEFINE) syntax string"""
+
+        output = ""
+        _dl = ["defined({})".format(n) if n[0] != "!" \
```
gilles-peskine-arm, line 123: Again, please don't use underscore-prefixed local variables. `dl` is acceptable, although a meaningful name would be better (I can't figure out what “dl” is abbreviating. Something list maybe? The reader shouldn't have to guess.).

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+        'psa_drv_slot_number_t': 'unsigned long long',
+        'psa_key_derivation_step_t': 'int',
+        'psa_key_id_t': 'unsigned',
+        'psa_key_slot_number_t': 'unsigned long long',
+        'psa_key_lifetime_t': 'unsigned',
+        'psa_key_type_t': 'unsigned',
+        'psa_key_usage_flags_t': 'unsigned',
+        'psa_pake_role_t': 'int',
+        'psa_pake_step_t': 'int',
+        'psa_status_t': 'int'
```
gilles-peskine-arm, line 239:
```suggestion
        'psa_status_t': 'int',
```

https://stackoverflow.com/questions/11597901/why-are-trailing-commas-allowed-in-a-list

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+import os
+from typing import Any, Iterator, List, Dict, Collection, Optional, Tuple
+
+from mbedtls_framework import build_tree
+from mbedtls_framework import c_parsing_helper
+from mbedtls_framework import c_wrapper_generator
+from mbedtls_framework import typing_util
+
+from mbedtls_framework.code_wrapper.psa_buffer import BufferParameter
+from textwrap import dedent
```
gilles-peskine-arm, line 19: Please group standard library imports, per [PEP 8](https://peps.python.org/pep-0008/#imports). (Pylint will nag you anyway, once you run it.) Also, PEP 8 doesn't require it, but please keep imports in the same group in module alphabetical order, that makes it easier to find something by scanning a list, so `textwrap` goes between `os` and `typing`.

scripts/mbedtls_framework/code_wrapper/psa_test_wrapper.py
gilles-peskine-arm: I'd prefer to have an unimportant newline disappear here than have this bizarre newline in an include directive argument.

gilles-peskine-arm (changes requested):
The new code generates identical output, which is good, but in several places, it's structured or written in ways that are hard to follow or maintain. That includes several places where you changed something that follows common conventions into something that doesn't (e.g. PEP 8, `main` function, commas at the end of multiline lists) and I don't understand why.

Also, pylint hasn't run on the new files.

scripts/mbedtls_framework/code_wrapper/psa_buffer.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class BufferParameter:
+    """Description of an input or output buffer parameter sequence to a PSA function."""
+    #pylint: disable=too-few-public-methods
+
+    def __init__(self, i: int, is_output: bool,
+                 buffer_name: str, size_name: str) -> None:
+        """Initialize the parameter information.
+        i is the index of the function argument that is the pointer to the buffer.
```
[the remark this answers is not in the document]
minosgalanakis: Apoogies, that was probably caused by a mistake in rebase conflict resolution. It was re-introduced in a consequtive [commit ](https://github.com/Mbed-TLS/mbedtls-framework/blob/08d5b60150511d8733a63ecec55f1da5def9fd01/scripts/mbedtls_framework/code_wrapper/psa_buffer.py)

scripts/generate_psa_wrappers.py
```diff
@@ -5,254 +5,15 @@
...
-        return super()._printf_parameters(typ, var)
-
+from mbedtls_framework.code_wrapper.psa_test_wrapper import PSATestWrapper, PSALoggingTestWrapper
 
 DEFAULT_C_OUTPUT_FILE_NAME = 'tests/src/psa_test_wrappers.c'
 DEFAULT_H_OUTPUT_FILE_NAME = 'tests/include/test/psa_test_wrappers.h'
 
-def main() -> None:
-    parser = argparse.ArgumentParser(description=globals()['__doc__'])
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
minosgalanakis: There are arguably some reasons to have a separate main() method but I would argue it not Pythonic. 

The whole scope of the automatic `__name__=="main"` check is to ensure that everything under it is not imported or executed other than by calling the module directly. Adding the requirement for main code to be importable, is a bit counter intuitive.

In this specific file all this code does is it import `argparser`, get the user input and call the appropriate libary methods to do what the user calling the script directly, wants to do. I fail to see how that code would/should be shared and exposed to other modules.

In addition `main()` would be evaulated on an import statement causing argparser to be needlesly imported.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -129,3 +152,17 @@ def read_function_declarations(functions: Dict[str, FunctionInfo],
...
                                        arguments)
+
+_C_TYPEDEF_DECLARATION_RE = r'typedef (?:struct )?(?P<type>\w+) (?P<name>\w+)'
+
+def read_typedefs(filename: str)->  Dict[str, str]:
+    """ Extract type definitions in a {typedef aliased name: original type} dictionary.
+    Multi-line typedef struct are not captured. """
+
+    type_decl = {}
+    with open(filename, encoding='utf-8') as inp:
```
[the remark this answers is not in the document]
minosgalanakis: I will change that so we can move on, but for the record I think that `read_logical_lines()` is hardly needed.

What it effective is doing is:

1. file.read()
2. re.sub
3. splitlines()
4. enumerate 
5. Iterate through the levels.

What is the method trying to achieve?
- If the ask is to skip comment lines, we can achieve that with rejecting regex, which is why the tool is build and tested for.
- If the ask is to find all matches in a multiline block of text, and return the line number, you can achieve it by calling re.match() to map the indexes and then re.finditer() once to have an iterator giving you the matches, that you can pair with the line number.

Why do we need custom methods for built and tested native python functionality?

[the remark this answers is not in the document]
gilles-peskine-arm: `read_logical_line` skips C comments. There's no native Python functionality for that.

> If the ask is to skip comment lines, we can achieve that with rejecting regex, which is why the tool is build and tested for.
>
> If the ask is to find all matches in a multiline block of text, and return the line number, you can achieve it by calling re.match() to map the indexes and then re.finditer() once to have an iterator giving you the matches, that you can pair with the line number.

Now you're discussing the implementation of `read_logical_lines`. I think you're outlining a different implementation technique. Rewriting `read_logical_lines` is out of scope here since we don't need any new functionality. Furthermore, your proposed technique seems more difficult than the one that's already there: it's hard to do regexp-based replacements but keep track of line numbers.

scripts/generate_psa_wrappers.py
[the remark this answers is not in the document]
gilles-peskine-arm: > There are arguably some reasons to have a separate main() method but I would argue it not Pythonic.

The official documentation says otherwise.

> Putting as few statements as possible in the block below if `__name__ == '__main__'` can improve code clarity and correctness. Most often, a function named main encapsulates the program’s primary behavior:

> In addition main() would be evaulated on an import statement causing argparser to be needlesly imported.

Uh? No, `main` is not evaluated on import. If the module is imported, either way, the same code gets executed. All the imports are at the top and unconditional, which is generally recommended unless there's a strong reason not to do it, which there isn't here.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
minosgalanakis: I was trying to be more helpfull to people reviewing the code. 

read_function_declarations is written as c-code with input and output parameter rather than return.

    def read_function_declarations(functions: Dict[str, FunctionInfo],
                                   filename: str) -> None:

People would expect it to be like:

    def read_function_declarations(filename: str) -> [None](functions: Dict[str, FunctionInfo]):

So this comment was meant to explain it.  I could word it like that? 

> """Reads functions to be wrapped from source header files into self.functions."""

gilles-peskine-arm: This comment is on `read_headers`, not on `read_function_declarations` (which hasn't changed).

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
minosgalanakis: Would that make it more readable? 

        _dl = ["defined({})".format(n) if n[0] != "!" \
               else "!defined({})".format(n[1:]) \
               for n in def_list]

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
minosgalanakis: It was meant for define_line. But since its an function internal variable I will keep it as `dl`

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
minosgalanakis: Good catch. Will add it.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
minosgalanakis: Will reword  it to:

`"""Configure instance's parameters from a user provided config."""`

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
gilles-peskine-arm: I think there's a misunderstanding here: I was talking about the output generated by the code, not about the presentation of the code. I would prefer to remove the part of the code that groups two parameters per line, because it would both make the generated C code easier to read and it would simplify the Python code that does the generation.

I didn't comment on the calculation of `_dl`, but I do prefer to have the `for` part on a separate line if the whole thing doesn't fit on a single line. Generally speaking, it's more readable if line breaks are at the loosest precedence levels, e.g. if `a + b * c` needs to be broken then it should be broken as `a + BREAK b * c` and not `a + b BREAK * c`. This is how I'd break down this line:
```
        _dl = [("defined({})".format(n) if n[0] != "!" else
                "!defined({})".format(n[1:]))
               for n in def_list]
```
Although it would probably be more readable to make the "[!]FOO" → "[!]defined(FOO)" transformation into an auxiliary function.

scripts/mbedtls_framework/code_wrapper/psa_buffer.py
```diff
@@ -14,13 +15,42 @@ class BufferParameter:
...
+
+        Write poisoning code if poison is true, unpoisoning code otherwise.
+        """
+        out.write('    MBEDTLS_TEST_MEMORY_{}({}, {});\n'.format(
+            'POISON' if poison else 'UNPOISON',
+            self.buffer_name, self.size_name
+        ))
+
+    @staticmethod
+    def poison_multi_write(out: typing_util.Writable,
```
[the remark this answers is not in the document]
minosgalanakis: This is tricky. 

The point of a data class is that it has knowlege of the data that it contains, and should be able to manipulate or present it in a custom format, without having to expose them to the calling one.

Otherwise we end up splitting the class's data manipulation accross two classes which is not ideal

I will propose the following:

1. Keep the text formatting as class function.
2. Set the indentation as an input parameter
3. Move the out.write() logic to the PSA_Wrapper.

e.g 

      def poison_wrap(self, poison: bool, ident_lv = "1") -> None:
          """Return a POISON/UNPOISON pr.
  
          return "{}MBEDTLS_TEST_MEMORY_{}({}, {});\n".format((ident_lv * 4) * " ",
                                                              POISON' if poison else 'UNPOISON',
                                                              self.buffer_name, self.size_name),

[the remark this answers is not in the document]
gilles-peskine-arm: > The point of a data class is that it has knowlege of the data that it contains, and should be able to manipulate or present it in a custom format, without having to expose them to the calling one.

Not necessarily. The `list` class doesn't have any knowledge of how to present the data that it contains. Some classes are collections whose job is to hold data however it's told; such classes allow the data to be accessed freely. Other classes have a job that includes protecting the integrity of the data; such classes only present formatted views of the data and only allow selected modifications. This is not a binary classification: many classes fall somewhere on the spectrum, possibly at different places for different parts of the data.

In this case, I prefer the way I did it originally. Keep `BufferParameter` as a data-oriented class that knows how to store the data, but not how to process it. This follows the general software engineering principle of keeping concerns separate (also known as single-responsibility principle and various other names). More concretely, here, it means that the wrapper generator classes keep full control as to how to format the C code, so maintainers don't have to worry about how a completely separate class formats C code. Conversely, `BufferParameter` might be used in contexts other than generating C code (for example to generate test cases in some systematic way), and in those contexts knowledge of C code formatting would be irrelevant, which shows that it isn't the job of `BufferParameter` to worry about C code formatting.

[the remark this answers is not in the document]
minosgalanakis: Also why shouldn't multi-write be a utliity method of the module instead of a statit method?

It needs to do 3 things:

- Check if the iterable parameters has any entries
- If it has add the `#if !defined(MBEDTLS_PSA_ASSUME_EXCLUSIVE_BUFFERS` guards
- If it has parameters, iterate through them and write them.

Moving the logic back to the psa_wrapper without that utility,  looks like that:


        if buffer_parameters:
            out.write('#if !defined(MBEDTLS_PSA_ASSUME_EXCLUSIVE_BUFFERS)\n')
            for param in buffer_parameters:
                out.write(param.poison_wrap(True))
            out.write('#endif /* !defined(MBEDTLS_PSA_ASSUME_EXCLUSIVE_BUFFERS) */\n')

        super()._write_function_call(out, function, argument_names)

        if buffer_parameters:
            out.write('#if !defined(MBEDTLS_PSA_ASSUME_EXCLUSIVE_BUFFERS)\n')
            for param in buffer_parameters:
                out.write(param.poison_wrap(False))
            out.write('#endif /* !defined(MBEDTLS_PSA_ASSUME_EXCLUSIVE_BUFFERS) */\n')

Which looks unnessary more redudant compared to:

        BufferParameter.poison_multi_write(out, buffer_parameters, True)
        super()._write_function_call(out, function, argument_names)
        BufferParameter.poison_multi_write(out, buffer_parameters, False)

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
minosgalanakis: I was planning on allowing in_headers being provided with an empty config. The check will be changed to `if in_headers and not config:`

scripts/mbedtls_framework/c_parsing_helper.py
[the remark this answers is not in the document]
minosgalanakis: > read_logical_line skips C comments. There's no native Python functionality for that.

I though regex exists for precicely that use-case?

Something like that should work 

    c_comments_rex = re.compile(r'//.*?$|/\*.*?\*/|\'(?:\\.|[^\\\'])*\'|"(?:\\.|[^\\"])*"', re.DOTALL | re.MULTILINE)

    # Remove comment lines
    content = re.sub(c_comments_rex, "", content)

    # Find typedefs
    for m in  re.finditer(_C_TYPEDEF_DECLARATION_RE, content):
        type_decl[m.group("name")] = m.group("type")

But again I am not exactly clear why `read_logical_lines` is needeed and what requirements it fulfills.  

Anyway this is out of scope so I am adjusting the utility function to use `read_logical_lines` so we can move forward.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,257 @@
...
+
+DEFAULTS = {
```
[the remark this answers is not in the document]
minosgalanakis: I have changed that but I would like to note my objection.

This is an immutable key:value store acting as a controlled user input. It does not need to be modified in any way. Dictionary is the best and more efficient way to handle it. 

What benefit are we getting by wrapping the said dictionary onto a Data-Class, other adding complexity?

### Dict design
~~~~~~

config = {key1: value1
          key2: value2
          ....}
          
def some_method(config: Dict[str, Any]= config) -> None:
~~~~~~ 

Here we have a global key:value string store, where the method takes the pointer as an argument.
The data is not modified in any way, and used when needed using that pointer.

### Proposed design
~~~~~~
class CFG.__dict__ =  {key1: value1
                       key2: value2
                       ....}
def some_method(config: CFG = CFG()) -> None:
~~~~~~

each time the some_method() is called, the following things happen.

- class CFG is memcopied onto a new location.
- The pointer of the copy is passed as the default argument of some_method.
- The pointer is added to the Garbage collector pool.
- Later on the garbage collector has to poll if that is still active and cleanup.

I fail to see why the second is prefered for configuration data that does not need to be manipulated, and do not require any special handling to be formatted. The only benefit I see is that it passes the contrains we set to ourselves by the limits of mypi.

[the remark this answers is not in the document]
gilles-peskine-arm: > This is an immutable key:value store acting as a controlled user input.

Not just that, **the set of keys is always the same**. That goes far from the basic properties of a dict, which have a variable and mutable set of keys. Python has a data structure exactly for that: a [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple). A namedtuple works well as long as you only want to bundle data together, without defining any special behavior on the class. If you want additional methods, a custom class tends to be simpler than extending the namedtuple.

> each time the some_method() is called, the following things happen.
>
> * class CFG is memcopied onto a new location.

No. The argument is passed by reference. Python doesn't have a concept of passing by value. Python implementations can pass objects by value _if_ the object is immutable (otherwise that would violate the semantics) and they decide that it's an optimization; I believe CPython does that for small integers. But in any case, a namedtuple or a class that's defined manually with similar properties would not be copied in any way that's different from a dict here.

> I fail to see why the second is prefered for configuration data that does not need to be manipulated

Not that this matters anyway, since we don't have particular performance or memory usage concerns in this script. The primary concern is readability and maintainability (the two are closely related). And in this respect, a namedtuple or similar is a clear win:

* A known set of elements, so it's easy for readers to understand what field names are valid, and static checkers can detect mistakes with field names.
* Each field's value has its own type of element. When fields are fields, it's easy to understand: `foo.bar` is an integer, `foo.qux` is a list of strings, etc. With dictionaries, you'd see `foo[field_name]` and you'd have no idea what that is.
* Immutable data structure, so no risk of modifying when unintended. (But even if the structure was mutable, a [`SimpleNamespace`](https://docs.python.org/3/library/types.html#types.SimpleNamespace) or a custom class would be preferable to a dictionary when the set of fields is always the same.)

> The only benefit I see is that it passes the contrains we set to ourselves by the limits of mypi.

You're switching cause and effect here. The data structure follows some rules, which are completely hidden if you stuff it in a `dict`. When you don't hide the rules, mypi helps you. When you hide the rules, you run into trouble with mypi: mypi is rightfully telling you that you're doing something wrong.

gilles-peskine-arm (changes requested):
Some issues are not resolved.

I'm not sure whether the commit structure is ok, for two reasons. Some commits are hard to understand because they do too many things. I've reviewed them but I was on the fence and I would understand it if the other reviewer requests restructuring. Furthermore there is a lot of back-and-forth that makes the code history harder to understand, especially since it's not just a commit and its reversal but several changes grouped in different ways with a net result that is still a change, just less drastic.

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -65,42 +67,50 @@ def _write_prologue(self, out: typing_util.Writable, header: bool) -> None:
...
         """
-        out.write("""/* Automatically generated by {}, do not edit! */
+        prologue = textwrap.dedent("""
+            /* Automatically generated by generate_psa_wrappers.py, do not edit! */
+
+            /* Copyright The Mbed TLS Contributors
+             * SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+             */
+
+        """).lstrip()
```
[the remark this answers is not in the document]
minosgalanakis: > Also, why complicate things with lstrip?

That allows more flexibility on how the strings are provided.

```
a = """ /* Automatically generated by generate_psa_wrappers.py, do not edit! */

           /* Copyright The Mbed TLS Contributors
""""
```

will produce the same result as:

```
a = """
            /* Automatically generated by generate_psa_wrappers.py, do not edit! */

            /* Copyright The Mbed TLS Contributors
"""
```

In more cases we may want to add empty newlines by appending at the end of the string, but it is not very usefull to append at the beggining of it.

> Also, I don't like using dedent: it only works if at least one line is indented at column 0, and it makes the generated snippets harder to read, especially if we want to use 80 columns of text

Personal preference aside,  [PEP style guide](https://peps.python.org/pep-0008/) recommends implicit concatenation as the way of handling multiline string lines. 

> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.
textwrap.dedent and inspect.cleandoc has been consis

For those who need other functionality textwrapt module has been the canonical way of handling it. Using mutlinline docutils breaking class identation hyerarcy is a non-cannonical approach, and we want the code to be maintained by people who are not familiar with it.

The limitations of dedent, do not apply in our use case since we are adding code at line 0. If something special is needed we can extend that function, but that is out of scope right now.

I will be adding a `blockstr_format` utiility function so we can update only that segment of code if needed.

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -65,42 +67,50 @@ def _write_prologue(self, out: typing_util.Writable, header: bool) -> None:
...
+
+            """.format(guard=self.header_guard)).lstrip()
+
+        for include in self._INCLUDES:
+            prologue += "#include {}\n".format(include)
+
+        # Make certain there is an empty line at the end of this section.
+        for i in [-1, -2]:
+            if prologue[i] != '\n':
+                prologue += ('\n')
```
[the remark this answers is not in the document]
minosgalanakis: Will replace it with:

     prologue += '\n' if self._INCLUDES else '\n\n'

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -65,42 +67,50 @@ def _write_prologue(self, out: typing_util.Writable, header: bool) -> None:
 
         This includes a description comment and some include directives.
         """
-        out.write("""/* Automatically generated by {}, do not edit! */
+        prologue = textwrap.dedent("""
```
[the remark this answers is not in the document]
minosgalanakis: We won't be stressing it out so it's not making a big difference, but the idea is that unless the file is used for log purposes, there it is better to open the file handler when we need to write and not keep it open when formatting the code.

Also one of the benefits for people debugging the code is that you can get to print the string, before it even reaches the point of generating the file.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,283 @@
...
+            return os.path.join(self.mbedtls_root, *path_list, filename)
+
+        return os.path.join(self.mbedtls_root, *path_list, filename)
+
+    # Utility Methods
+    @staticmethod
+    def parse_def_guards(def_list: Collection[str])-> str:
+        """ Create define guards.
+
+            Parses an input list of format [[[C] preprocessor] macro, ...] and
```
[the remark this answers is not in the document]
minosgalanakis: This will be reworded

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -157,6 +155,27 @@ def _parameter_should_be_copied(function_name: str,
 
         return True
 
+    def _poison_wrap(self, param : BufferParameter, poison: bool, ident_lv = 1) -> str:
+        """Returns a custom string based on the values of param and poison."""
```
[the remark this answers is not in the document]
minosgalanakis: This will be reworded

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,283 @@
...
+        self.read_config(config)
+
+        if in_headers and not config:
+            self.read_headers(in_headers)
+
+    def read_config(self, cfg: PSAWrapperCFG)-> None:
+        """Configure instance's parameters from a user provided config."""
+        if not cfg:
+            return
+        self._CPP_GUARDS = PSAWrapper.parse_def_guards(cfg.define_guards)
```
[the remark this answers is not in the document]
minosgalanakis: Which is the private file here? neither `self._CPP_GUARDS` nor  `cfg.define_guards` fit that description?

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -16,23 +16,25 @@
 
 from .psa_buffer import BufferParameter
 
-DEFAULTS = {
-    "input_headers" : ['crypto.h', 'crypto_extra.h'],
-    "define_guards" : ["MBEDTLS_PSA_CRYPTO_C", "MBEDTLS_TEST_HOOKS", "!RECORD_PSA_STATUS_COVERAGE_LOG"],
-    "skip_list" : frozenset([
+class PSAWrapperCFG:
```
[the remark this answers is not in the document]
minosgalanakis: check-python-files.sh has been updated by another pr already.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,283 @@
...
+
+from .. import build_tree
+from .. import c_parsing_helper
+from .. import c_wrapper_generator
+from .. import typing_util
+
+from .psa_buffer import BufferParameter
+
+class PSAWrapperCFG:
+    input_headers = ['crypto.h', 'crypto_extra.h']
```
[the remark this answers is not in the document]
minosgalanakis: Because the class was originally a dictionary and was considered [bad implementation](https://github.com/Mbed-TLS/mbedtls-framework/pull/25#discussion_r1654996924) in the previous round of reviews. 

And given that is a dataclass there is no need to consider instance variables. When we override, the members tha are outdates should be overriden.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -16,23 +16,25 @@
 
 from .psa_buffer import BufferParameter
 
-DEFAULTS = {
-    "input_headers" : ['crypto.h', 'crypto_extra.h'],
-    "define_guards" : ["MBEDTLS_PSA_CRYPTO_C", "MBEDTLS_TEST_HOOKS", "!RECORD_PSA_STATUS_COVERAGE_LOG"],
-    "skip_list" : frozenset([
+class PSAWrapperCFG:
+    input_headers = ['crypto.h', 'crypto_extra.h']
```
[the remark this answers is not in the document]
minosgalanakis: I have honestly considered the same.  I can remove them from the configuration and return them as optional PSAWrapper class variables, that can be overloaded by user upon instantiation. 

If in the future we want to migrate them back to config data-class we can.

scripts/generate_psa_wrappers.py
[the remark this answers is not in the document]
minosgalanakis: That won't be changed. There is practically no reason for invoking argparser on a separate main function. 

We should not be aiming to complicate code for the sake of it.

`main()` has purpose in complicated multithreaded programs, like async servers, games etc. Not for a script passing two parameters onto single line of code.

scripts/mbedtls_framework/c_wrapper_generator.py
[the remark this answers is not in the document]
gilles-peskine-arm: > That allows more flexibility on how the strings are provided.

I don't see how that's a good thing. It allows more flexibility to make the source code less readable. Rather a bad thing in my view.

> PEP style guide recommends implicit concatenation as the way of handling multiline string lines.

That's about long lines of Python code. PEP 8 doesn't say anything about long string literals.

> For those who need other functionality textwrapt module has been the canonical way of handling it.

I don't know what other functionality you're referring to, but `textwrap` is mostly about text wrapping and we don't want text wrapping here.

> Using mutlinline docutils breaking class identation hyerarcy is a non-cannonical approach, and we want the code to be maintained by people who are not familiar with it.

I don't know what you would consider to be the canonical approach here. And I don't see what's hard to maintain about having string literals that are used as presented in the source code, as opposed to being munged by some function that readers may be unfamiliar with.

> The limitations of dedent, do not apply in our use case since we are adding code at line 0.

ITYM at column 0. We are not only adding code at column 0: we often add code inside functions.

So my objection still stands: you're making the code less readable.

scripts/mbedtls_framework/c_parsing_helper.py
```diff
@@ -82,7 +82,7 @@ def __str__(self) -> str:
         str_text = "{} {} {}({})".format(" ".join(self.qualifiers),
```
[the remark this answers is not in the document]
gilles-peskine-arm: Please make one commit per topic. The fact that multiple issues were raised in the same review is not at all a reason to group them in the same commit.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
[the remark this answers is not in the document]
gilles-peskine-arm: Here the point is _lowercase_ instead of uppercase. The name should be `self.cpp_guards` to convey that it's ok to access the field from outside the class, or `self._cpp_guards` to convey that only methods of the class or a descendant should access the field.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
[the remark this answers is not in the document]
gilles-peskine-arm: > given that is a dataclass there is no need to consider instance variables.

I can't figure out why something being a “dataclass” has anything to do with considering instance variables. I'm not even sure what you call a dataclass. This isn't what [Python itself calls a dataclass](https://docs.python.org/3/library/dataclasses.html).

scripts/generate_psa_wrappers.py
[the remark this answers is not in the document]
gilles-peskine-arm: Sorry, but that's a blocker. A `main` function is [the recommended way to structure non-toy programs](https://docs.python.org/3/library/__main__.html#idiomatic-usage) and has objective advantages as explained above.

Even pylint will reject your code because `parser` and other variables are global variables and it will want them to use the naming convention for global variables. Don't disable the warning: fix the code.

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -65,42 +70,49 @@ def _write_prologue(self, out: typing_util.Writable, header: bool) -> None:
 
         This includes a description comment and some include directives.
         """
-        out.write("""/* Automatically generated by {}, do not edit! */
+        prologue = blockstr_format(f'''
```
gilles-peskine-arm, line 73: Continuing https://github.com/Mbed-TLS/mbedtls-framework/pull/25#discussion_r1672223530 — This deintentation approach will fail as soon as we start generating multiline blocks of code inside functions. So it's not future proof and I would prefer not to use it. This is an objective argument, in addition to the subjective argument about readability.

I'm not going to block it because I want to move on. I do think it is a waste of time that this pull request makes a change that we're going to undo later, but I'm not going to insist on reverting that change now.

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -25,12 +25,15 @@ def c_declare(prefix: str, name: str, suffix: str) -> str:
...
 
-WrapperInfo = typing.NamedTuple('WrapperInfo', [
+WrapperInfo = NamedTuple('WrapperInfo', [
     ('argument_names', List[str]),
     ('guard', Optional[str]),
     ('wrapper_name', str),
 ])
 
+def blockstr_format(in_str: str, new_lines: int = 1) -> str:
+    """Return a whitespace stripped str, appended with user defined empty lines."""
```
gilles-peskine-arm, line 35: The description omits the important aspect of this function, which is that it removes indentation! Please change the one-line description to focus on the indentation removal. In the long description, mention stripping leading and trailing empty lines, the effect of `new_lines`, and a warning that this function is only suitable if the result is supposed to be vertically aligned to column 0.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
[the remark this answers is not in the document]
gilles-peskine-arm: This is still not addressed as of 03bfc58000f7e6ff35a0f4a97821619c054ee275.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,286 @@
...
+
+from .. import build_tree
+from .. import c_parsing_helper
+from .. import c_wrapper_generator
+from .. import typing_util
+
+from .psa_buffer import BufferParameter
+
+class PSAWrapperConfiguration:
+    """Configuration dataclass for PSA Wrapper."""
```
gilles-peskine-arm, line 20: This isn't what Python calls a “dataclass”, so please don't use that technical term. The generic expression “data class” would be accurate.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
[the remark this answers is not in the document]
gilles-peskine-arm: This is not addressed as of 03bfc58000f7e6ff35a0f4a97821619c054ee275.

scripts/generate_psa_wrappers.py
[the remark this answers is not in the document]
gilles-peskine-arm: This is not addressed as of 03bfc58000f7e6ff35a0f4a97821619c054ee275.

gilles-peskine-arm (changes requested):
Several problems are still not addressed.

As a general remark, when a static checker tells you something is wrong, think about what is wrong instead of blindly silencing the warning.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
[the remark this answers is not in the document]
minosgalanakis: Can you elaborate on how that should be addressed?

scripts/generate_psa_wrappers.py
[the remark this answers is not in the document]
minosgalanakis: This is the definiton of a toy program. We are just passing around some parameters from command line into a class. 

I will ammend it as you requested, but I am not confident that we should be letting the tools dictate the complexity of our code.

scripts/mbedtls_framework/c_wrapper_generator.py
minosgalanakis: I have rewritten the implementation with a regex single liner which will satisfy that future requirement.

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -31,9 +30,16 @@ def c_declare(prefix: str, name: str, suffix: str) -> str:
...
 
-def blockstr_format(in_str: str, new_lines: int = 1) -> str:
-    """Return a whitespace stripped str, appended with user defined empty lines."""
-    return inspect.cleandoc(in_str) + ("\n" * (new_lines + 1))
+def strip_indentation(in_str: str, new_lines: int = 1, ident_lv: int = 4) -> str:
+    """Return a whitespace stripped str, appended with user defined empty lines.
+
+    The method will remove (ident_lv * 4) space-character identations from
+    input string. It will also remove all whitespace around the text-block.
+    At the end of the stream a `new_lines` ammount of empty lines will be added.
```
gilles-peskine-arm, line 38:
```suggestion
    At the end of the string a `new_lines` amount of empty lines will be added.
```

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -31,9 +30,16 @@ def c_declare(prefix: str, name: str, suffix: str) -> str:
     ('wrapper_name', str),
 ])
 
-def blockstr_format(in_str: str, new_lines: int = 1) -> str:
-    """Return a whitespace stripped str, appended with user defined empty lines."""
-    return inspect.cleandoc(in_str) + ("\n" * (new_lines + 1))
+def strip_indentation(in_str: str, new_lines: int = 1, ident_lv: int = 4) -> str:
```
gilles-peskine-arm, line 33: Why is the default `indent_lv` 4? Also, I'm not sure it makes sense to specify how much indentation to strip, given that the intent is that the input's indentation is based on the indentation of the surrounding code, while the output's indentation is based on the position in the generated code. The input's indentation can be determined by stripping the common whitespace sequence from the beginning of lines (line `dedent` and `cleandoc`). Let the user specify the desired output indentation: 0 for code outside a function, 1 for code in a function, 2 for code in a block in a function, etc.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,286 @@
...
+        self.read_config(config)
+        self.read_headers(in_headers)
+
+    def read_config(self, cfg: PSAWrapperConfiguration)-> None:
+        """Configure instance's parameters from a user provided config."""
+
+        self._CPP_GUARDS = PSAWrapper.parse_def_guards(cfg.cpp_guards)
+        self._SKIP_FUNCTIONS = cfg.skipped_functions
+        self._FUNCTION_GUARDS.update(cfg.function_guards)
+        self._NOT_IMPLEMENTED = cfg.skipped_argument_types
```
gilles-peskine-arm, line 92: As [previously discussed](https://github.com/Mbed-TLS/mbedtls-framework/pull/25#discussion_r1672399620), these are instance variables, not class variables, so the [naming convention](https://peps.python.org/pep-0008/#method-names-and-instance-variables) is lowercase. A leading underscore indicates that code outside the class should not access these fields (which I'm indifferent about).

scripts/generate_psa_wrappers.py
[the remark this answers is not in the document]
gilles-peskine-arm: This is not a toy program: it's hundreds of lines of code divided in several files. The “tools” in question are the language itself. Pylint isn't always right, but in this case, it is correctly pointing out some dodgy code.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
```diff
@@ -0,0 +1,286 @@
...
+from .. import c_parsing_helper
+from .. import c_wrapper_generator
+from .. import typing_util
+
+from .psa_buffer import BufferParameter
+
+class PSAWrapperConfiguration:
+    """Configuration data class for PSA Wrapper."""
+
+    cpp_guards = ["MBEDTLS_PSA_CRYPTO_C", "MBEDTLS_TEST_HOOKS", "!RECORD_PSA_STATUS_COVERAGE_LOG"]
```
gilles-peskine-arm, line 22: (Resuming https://github.com/Mbed-TLS/mbedtls-framework/pull/25#discussion_r1672417045)

> > These all should be instance variables, not class variables. Otherwise we will be very confused when we instantiate or override the class.
> >
> > Why isn't the class a [NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple)?
>
> Can you elaborate on how that should be addressed?

Either use a `NamedTuple`, or make it a data class, as in, a class whose sole purpose is to store data into fields — what C would call a struct. Pylint doesn't like data classes, but it doesn't have a very strong justification. It wants to drive you to the mechanisms for building data classes through the standard library, but those mechanisms aren't particularly useful unless you want to do some introspection, which we don't care about here.

A custom data class would be just a constructor that initializes the fields to their default value.

```
class PSAWrapperConfiguration:
    """Configuration data class for PSA Wrapper."""
    def __init__(self) -> None:
        self.cpp_guards = ["MBEDTLS_PSA_CRYPTO_C", "MBEDTLS_TEST_HOOKS", "!RECORD_PSA_STATUS_COVERAGE_LOG"]
        …
```

gilles-peskine-arm (changes requested):
We're getting close!

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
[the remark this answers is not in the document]
gilles-peskine-arm: (continuing at https://github.com/Mbed-TLS/mbedtls-framework/pull/25#discussion_r1682437015)

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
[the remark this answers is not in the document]
gilles-peskine-arm: (continuing at https://github.com/Mbed-TLS/mbedtls-framework/pull/25#discussion_r1682463049)

scripts/mbedtls_framework/c_wrapper_generator.py
minosgalanakis: > Why is the default indent_lv 4? 

Pure statistical coincident. We use it most with level 4.

I have considered making it read the first line of code and automatically determine the level in 4 spaces segments. But giving the control of the filter to the caller would allow creating snippets like that. ( which are addmittely not very usefull )

~~~~~
        some_function();
     some_less_indented_code();
~~~~~

Would that be acceptable?

- Automatically remove indentation.
- Change the indent_lv parameter to adjust the output indendation?

gilles-peskine-arm: > Automatically remove indentation.
> Change the indent_lv parameter to adjust the output indendation?

Yes, I think that's the best choice if we don't preserve the source indentation.

scripts/mbedtls_framework/code_wrapper/psa_wrapper.py
minosgalanakis: I am getting a bit of mixed signals, as to why we are going around in circles on this one.  The aim was to use a generic key:value configuration datastore, which allows other modules to import and .update() as required.

Then [as requested](https://github.com/Mbed-TLS/mbedtls-framework/pull/25#discussion_r1654996924) this was changed into a class because a using a dictionary is stringy typing? Now I am being asked to use a NamedTuple which completely fails the original intent, and is also stringy typing? 

Tuples are immutable, which means in order to update an entry the consumer will have to recreate it. There is a reason that dictionaries are universaly used for configuration input.

Now for dataclass argument, yes I understand that this is not a [dataclass](https://docs.python.org/3/library/dataclasses.html) as introduced in Python3.7, but the proposal is to basically implement in Python3.6 what the dataclass decorator does.

Is that all that for the shake of adding-complexity really needed?

gilles-peskine-arm: > The aim was to use a generic key:value configuration datastore, which allows other modules to import and .update() as required.

Uh? That's the first time this comes up. Why should this be a generic configuration datastore? Why would modules add their own configuration entries that the base class doesn't know about? That seems like a solution in need of a problem.

> a NamedTuple which completely fails the original intent, and is also stringy typing?

A NamedTuple isn't extensible. But extensibility is a new requirement and I don't see the point of this new requirement.

A NamedTuple is not stringly typed. Its type is the NamedTuple. The set of keys is known, and the type of the key corresponding to each value is known.

Note once again that I'm not saying you _have_ to use a NamedTuple. I'd be happy if you just made the change I posted above, changing the class fields into instance fields. With `# pylint: disable=too-few-public-methods` because Pylint is opinionated about not doing your own data class, but there isn't a strong objective justification behind that opinion.

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -207,15 +237,15 @@ def _skip_function(self, function: FunctionInfo) -> bool:
             return True
         return False
 
-    _FUNCTION_GUARDS = {
+    _function_guards = {
```
gilles-peskine-arm, line 240: I missed that earlier: since this is now mutable, it should be an instance field, set in the class's construtor, instead of a class field that's created when the class is defined.

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -30,16 +30,28 @@ def c_declare(prefix: str, name: str, suffix: str) -> str:
...
 ])
 
-def strip_indentation(in_str: str, new_lines: int = 1, ident_lv: int = 4) -> str:
-    """Return a whitespace stripped str, appended with user defined empty lines.
+def strip_indentation(in_str: str, new_lines: int = 1, ident_lv: int = 0) -> str:
+    """Return a whitespace stripped str, with configurable whitespace in output.
 
-    The method will remove (ident_lv * 4) space-character identations from
-    input string. It will also remove all whitespace around the text-block.
+    The method will remove space-character identations from input string.
```
gilles-peskine-arm, line 36: Given the following input:
```
  foo
   bar
 qux
```

The documentation says that indentation is stripped on each line, so the output is:
```
foo
bar
qux
```

The actual output strips the indentation of the first line with a non-zero indent (which is pretty bizarre):
```
foo
 bar
 qux
```

The intent is to strip the common indentation, so the output should be:
```
 foo
  bar
qux
```

Please fix. You can call `textwrap.dedent`.

gilles-peskine-arm (changes requested):
Almost there

scripts/mbedtls_framework/c_wrapper_generator.py
minosgalanakis: Since the method is already reading the identation levels `    _input_idents = [len(n) for n in re.findall(r'(?m)^ +', in_str)]` a call to `min(_input_idents )` will be sufficent to achieve that formatting. 

Will update

scripts/mbedtls_framework/c_wrapper_generator.py
```diff
@@ -25,12 +24,34 @@ def c_declare(prefix: str, name: str, suffix: str) -> str:
...
+    """
+
+    _ret_string = in_str.strip('\n')
+
+    # Count empty spaces in beggining of each line. The smallest non-zero entry
+    # will be used to clean up input indentation.
+    _common_indents = min([len(n) for n in re.findall(r'(?m)^ +', in_str)])
+    if _common_indents:
+        _ret_string = re.sub(r'(?m)^ {{{indent}}}'.format(indent=_common_indents),
+                             '', _ret_string)
```
gilles-peskine-arm, line 50: This errors out if no line is indented. E.g.
```suggestion
    indents = [len(n) for n in re.findall(r'(?m)^ +', in_str)]
    if indents:
        _ret_string = re.sub(r'(?m)^ {{{indent}}}'.format(indent=min(indents)),
                             '', _ret_string)
```

gilles-peskine-arm (changes requested):
LGTM except one edge case

gilles-peskine-arm (approved):
LGTM

valeriosetti (approved):

PR #27, opened 2024-06-12
Add sanitizer build option when building a file

gabor-mezei-arm:
Add the `-fsanitize` build option for C file building in the scripts.

When `MBEDTLS_TEST_CONSTANT_FLOW_MEMSAN` is set, the `check_config.h` have a check for the memory sanitizer is enabled in the compiler. This can cause build error when reading the configuration from the scripts.

### DIFF
--- a/scripts/mbedtls_framework/c_build_helper.py
+++ b/scripts/mbedtls_framework/c_build_helper.py
@@ -7,6 +7,7 @@
 
 import os
 import platform
+import re
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (commented):
Originally `c_build_helper` was only used as part of `make generated_files`, in scripts that expect the original configuration but also work if a different set of options are commented out. I wonder if we should focus on that objective or instead make the code more general. This pull request makes the code more general, but I wonder if this is the right direction.

One thing that's been bothering me (and some users) is that this script assumes a native C compiler, but honors `CC` and now `CFLAGS`. It was a deliberate choice not to honor `CFLAGS` to make it work at least in some cases of cross compilation (if your `CC` is a multi-architecture compiler like `clang`, and it's `CFLAGS` that causes the compilation to be cross-compilation).

Maybe instead of fiddling with `CFLAGS`, we should instead act on the configuration, and try to ensure that the compilation runs in a “good” configuration?

How did this problem come up anyway? Trying to run `make generated_files` with a modified configuration? (That doesn't always work, which is why I almost always do my non-default-configuration builds by setting `MBEDTLS_CONFIG_FILE` or using an extra include directory containing `mbedtls/mbedtls_config.h`, instead of editing `include/mbedtls/mbedtls_config.h`.)

gabor-mezei-arm:
Not related any more. The problem is fixed with ensuring the generation of the concerned files before the build phase.

PR #29, opened 2024-06-20
Disable allow_abbrev from Python scripts using argparse

juhaylinen:
Python's argparse library, by default, allows shortening of command line arguments. This can introduce silent failures when shortened commands are used and another command is added to the script which uses that name.

This change was requested on https://github.com/Mbed-TLS/mbedtls/pull/9294


gilles-peskine-arm (dismissed):
LGTM

juhaylinen:
Rebased

PR #38, opened 2024-07-23
c_wrapper_generator: Increased robustness of whitespace capture.

minosgalanakis:
Minor fix, which makes strip_indentation() to only consider lines with non whitespace characters following spaces, when trying to determine the common identation.

eg. For a common copyright block like bellow:

~~~~~
'''
            /* Copyright The Mbed TLS Contributors                      -> Considered
             * SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later  -> Considered
             */                                                         -> Considered

    '''                                                                 -> Not Considered
~~~~~

The Common calculated whitespace will be 8, not 4


davidhorstmann-arm (approved):
LGTM, thanks!

paul-elliott-arm (approved):
LGTM

PR #42, opened 2024-08-22
Fix everest build issue

waleed-elmelegy-arm:
Fix everest build issue as a result of moving everest during repo split


davidhorstmann-arm (approved):
LGTM

gilles-peskine-arm (approved):
LGTM

PR #43, opened 2024-09-05
Update data_files/.gitignore with test output

mpg:
Before this commit, if I run 'make test' in an Mbed TLS work tree, then afterwards 'git status' lists the framework submodule as unclean (untracked files) which I find annoying.


davidhorstmann-arm (approved):
LGTM, thanks!

gilles-peskine-arm (approved):

PR #45, opened 2024-09-10
Projects and Mbed TLS 3.6 detection

ronald-cron-arm:
Changes to build_tree.py to use project_name.txt to detect between Mbed TLS and TF-PSA-Crypto and to detect an Mbed TLS 3.6 working tree.
Companion of https://github.com/Mbed-TLS/mbedtls/pull/9552.

### DIFF
--- a/scripts/mbedtls_framework/build_tree.py
+++ b/scripts/mbedtls_framework/build_tree.py
@@ -130,12 +130,15 @@ def guess_tf_psa_crypto_root(root: Optional[str] = None) -> str:
     else:
         raise Exception('TF-PSA-Crypto source tree not found')
 
-def is_mbedtls_3_6_branch(root: Optional[str] = None) -> bool:
-    """Whether root is the root directory of the Mbed TLS 3.6 branch """
-    if root is None:
-        root = os.getcwd()
+def is_mbedtls_3_6() -> bool:
+    """Whether the working tree is an Mbed TLS 3.6 one or not
+
+    Return false in we are in TF-PSA-Crypto or in Mbed TLS but with a version
+    different from 3.6.x.
+    Raise an exception if we are neither in Mbed TLS nor in TF-PSA-Crypto.
+    """
+    root = guess_project_root()
     if not looks_like_mbedtls_root(root):
         return False
     with open(os.path.join(root, 'include', 'mbedtls', 'build_info.h'), 'r') as f:
-        pattern = re.compile("#define MBEDTLS_VERSION_NUMBER.*0x0306")
-        return pattern.search(f.read()) is not None
+        return re.search(r"#define MBEDTLS_VERSION_NUMBER.*0x0306", f.read()) is not None
--- a/scripts/mbedtls_framework/build_tree.py
+++ b/scripts/mbedtls_framework/build_tree.py
@@ -11,13 +11,19 @@ from typing import Optional
 
 def looks_like_tf_psa_crypto_root(path: str) -> bool:
     """Whether the given directory looks like the root of the PSA Crypto source tree."""
-    return all(os.path.isdir(os.path.join(path, subdir))
-               for subdir in ['include', 'core', 'drivers', 'programs', 'tests'])
+    try:
+        with open(os.path.join(path, 'scripts', 'project_name.txt'), 'r') as f:
+            return f.read() == "TF-PSA-Crypto\n"
+    except FileNotFoundError:
+        return False
 
 def looks_like_mbedtls_root(path: str) -> bool:
     """Whether the given directory looks like the root of the Mbed TLS source tree."""
-    return all(os.path.isdir(os.path.join(path, subdir))
-               for subdir in ['include', 'library', 'programs', 'tests'])
+    try:
+        with open(os.path.join(path, 'scripts', 'project_name.txt'), 'r') as f:
+            return f.read() == "Mbed TLS\n"
+    except FileNotFoundError:
+        return False
 
 def looks_like_root(path: str) -> bool:
     return looks_like_tf_psa_crypto_root(path) or looks_like_mbedtls_root(path)
--- a/scripts/mbedtls_framework/build_tree.py
+++ b/scripts/mbedtls_framework/build_tree.py
@@ -7,6 +7,7 @@
 
 import os
 import inspect
+import re
 from typing import Optional
 
 def looks_like_tf_psa_crypto_root(path: str) -> bool:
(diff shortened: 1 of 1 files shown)


davidhorstmann-arm (dismissed):
LGTM pending other PR review, thanks!

scripts/mbedtls_framework/build_tree.py
```diff
@@ -7,17 +7,24 @@
...
 import inspect
+import re
 from typing import Optional
 
 def looks_like_tf_psa_crypto_root(path: str) -> bool:
     """Whether the given directory looks like the root of the PSA Crypto source tree."""
-    return all(os.path.isdir(os.path.join(path, subdir))
-               for subdir in ['include', 'core', 'drivers', 'programs', 'tests'])
+    try:
+        with open(os.path.join(path, 'scripts', 'project_name.txt'), 'r') as f:
```
minosgalanakis, line 16: I may have missed that discusion since I was away but is using a `project_name.txt` file the cannocnical and universal accepted way to convey infromation between the mbedtls branches?

Or is that a temporary workaround?

scripts/mbedtls_framework/build_tree.py
```diff
@@ -7,17 +7,24 @@
...
+    except FileNotFoundError:
+        return False
 
 def looks_like_mbedtls_root(path: str) -> bool:
     """Whether the given directory looks like the root of the Mbed TLS source tree."""
-    return all(os.path.isdir(os.path.join(path, subdir))
-               for subdir in ['include', 'library', 'programs', 'tests'])
+    try:
+        with open(os.path.join(path, 'scripts', 'project_name.txt'), 'r') as f:
+            return f.read() == "Mbed TLS\n"
```
minosgalanakis, line 25: It feels a bit redudant to have two identical methods for two different comparison operators. Also the name could be confusing since we are looking for the project rather than the root directory.

I would propose moving forward to have a uniform method:

    get_project_from_path(path:str)-> str: 

That returns the name of the project, or "" if invalid.

scripts/mbedtls_framework/build_tree.py
```diff
@@ -122,3 +129,13 @@ def guess_tf_psa_crypto_root(root: Optional[str] = None) -> str:
         return root
     else:
         raise Exception('TF-PSA-Crypto source tree not found')
+
+def is_mbedtls_3_6_branch(root: Optional[str] = None) -> bool:
+    """Whether root is the root directory of the Mbed TLS 3.6 branch """
+    if root is None:
+        root = os.getcwd()
```
minosgalanakis, line 136: why set it to none and check it in the body of the method while we could be setting the `os.getcwd()` as the default?

Also I would mention in the docstring that the method will try to check the current directory. 

I am personally not a fan of using getcdw() since that stops method from working when invoked by other directories and other scripts.

scripts/mbedtls_framework/build_tree.py
```diff
@@ -122,3 +129,13 @@ def guess_tf_psa_crypto_root(root: Optional[str] = None) -> str:
...
+    if not looks_like_mbedtls_root(root):
+        return False
+    with open(os.path.join(root, 'include', 'mbedtls', 'build_info.h'), 'r') as f:
```
minosgalanakis, line 139: There is no error handling or documentation of what will happen if the build_info cannot be found and open will throw an exception.

scripts/mbedtls_framework/build_tree.py
```diff
@@ -122,3 +129,13 @@ def guess_tf_psa_crypto_root(root: Optional[str] = None) -> str:
...
+        pattern = re.compile("#define MBEDTLS_VERSION_NUMBER.*0x0306")
```
minosgalanakis, line 140: If method is meant to return true if and only if:

1. We are in an MbedTLS root
2. Because we are in an MbedTLS root the path 'include/mbedtls/build_info.h` exists
3. The `MBEDTLS_VERSION_NUMBER == 0x0306`

Wouldn't it be simpler to write like that?

~~~~~
def is_mbedtls36_branch():
  try:
     with open(os.path.join(root, 'include', 'mbedtls', 'build_info.h'), 'r') as f:
       if re.findall(r"#define MBEDTLS_VERSION_NUMBER.*0x0306", f.read()):
           return True
   except:
       pass
   return False

~~~~~

minosgalanakis (commented):
It seems to do the job but feels a bit hackish. If that is an intermediate patch I am not insisting on changes but it feels like the helper functions are growing out of control.

If we agreed that we will be using a text file to communicate state between branches, all of those methods could be replaced by one, and the data could be stored in a universally accepted and easy to parse way like json in tha file.

scripts/mbedtls_framework/build_tree.py
ronald-cron-arm: `project_name.txt` has been introduced in #8609 to detect between Mbed TLS and TF-PSA-Crypto projects. I am not extending it here to be able to detect between branches of the projects because it would be yet another file to update when we change version.

scripts/mbedtls_framework/build_tree.py
ronald-cron-arm: Yes we have twice the same `open` line of code but only twice thus I do not feel the need to factorize it in a dedicated function.
Otherwise I think the name of the function are still fine, they still detect if `path` looks like the root of Mbed TLS or TF-PSA-Crypto, just they do it differently.

scripts/mbedtls_framework/build_tree.py
ronald-cron-arm: Yes this was intended. The `build_info.h` path is hardcoded in the function thus I prefer to stop immediately in this function if the path is not correct anymore.

scripts/mbedtls_framework/build_tree.py
ronald-cron-arm: I've removed the `root` parameter that I found eventually just confusing.

scripts/mbedtls_framework/build_tree.py
ronald-cron-arm: I've tried to improve the function scope, its documentation and implementation. Please have another look, thanks.

minosgalanakis (approved):
LGTM

ronald-cron-arm:
#9552 (dev) CI is green and #9553 (3.6) CI had a DTLS glitch but can be considered green I think. @davidhorstmann-arm @minosgalanakis please have another look.

davidhorstmann-arm (approved):
LGTM but noting that `is_mbedtls_3_6()` is not tested or used anywhere. This is okay because it is script-only, but I have tested locally on the heads of #9552 and #9553 to make sure

PR #55, opened 2024-10-03
Framework: Split check_test_cases.py and outcome_analysis.py

gilles-peskine-arm:
* Move `collect_test_cases.py` (split from `check_test_cases.py`), `check_test_cases.py` and `outcome_analysis.py` (split from `analyze_outcomes.py`) to the framework repository. Prior to the move, all files were made identical in 3.6 and development.
* By default, it is now an error if a test case is not executed, but the calling script can downgrade it back to a warning.
* Quality of life improvement for when outcome analysis fails: the log of outcome analysis will be the `outcome_analysis-analyze_outcomes.log.xz` artifact on the CI. That way you won't have to dig into the classic pipeline view to see the failures.
* Quality of life improvement for when outcome analysis fails: the existing artifact `outcomes.csv.xz` can now be used locally without decompressing it. (This requires a mypy upgrade, done in each consuming branch.)

* development: https://github.com/Mbed-TLS/mbedtls/pull/9668
* 3.6: https://github.com/Mbed-TLS/mbedtls/pull/9669

### DIFF
--- a/scripts/mbedtls_framework/outcome_analysis.py
+++ b/scripts/mbedtls_framework/outcome_analysis.py
@@ -9,6 +9,8 @@ the classes with branch-specific customizations such as ignore lists.
 # SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
 
 import argparse
+import gzip
+import lzma
 import sys
 import traceback
 import re
@@ -112,11 +114,19 @@ def name_matches_pattern(name: str, str_or_re: IgnoreEntry) -> bool:
     else:
         return str_or_re == name
 
+def open_outcome_file(outcome_file: str) -> typing.TextIO:
+    if outcome_file.endswith('.gz'):
+        return gzip.open(outcome_file, 'rt', encoding='utf-8')
+    elif outcome_file.endswith('.xz'):
+        return lzma.open(outcome_file, 'rt', encoding='utf-8')
+    else:
+        return open(outcome_file, 'r', encoding='utf-8')
+
 def read_outcome_file(outcome_file: str) -> Outcomes:
     """Parse an outcome file and return an outcome collection.
     """
     outcomes = {}
-    with open(outcome_file, 'r', encoding='utf-8') as input_file:
+    with open_outcome_file(outcome_file) as input_file:
         for line in input_file:
             (_platform, component, suite, case, result, _cause) = line.split(';')
             # Note that `component` is not unique. If a test case passes on Linux
@@ -301,7 +311,7 @@ def main(known_tasks: typing.Dict[str, typing.Type[Task]]) -> None:
     try:
         parser = argparse.ArgumentParser(description=__doc__)
         parser.add_argument('outcomes', metavar='OUTCOMES.CSV',
-                            help='Outcome file to analyze')
+                            help='Outcome file to analyze (can be .gz or .xz)')
         parser.add_argument('specified_tasks', default='all', nargs='?',
                             help='Analysis to be done. By default, run all tasks. '
                                  'With one or more TASK, run only those. '
--- a/scripts/mbedtls_framework/outcome_analysis.py
+++ b/scripts/mbedtls_framework/outcome_analysis.py
@@ -47,9 +47,20 @@ Outcomes = typing.Dict[str, ComponentOutcomes]
 class Results:
     """Process analysis results."""
 
-    def __init__(self):
+    def __init__(self,
+                 stderr: bool = True,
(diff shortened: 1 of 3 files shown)


eleuzi01 (approved):
LGTM

eleuzi01 (changes requested):
Having moved files to the framework myself, I now realize that this move does not preserve file history.

scripts/mbedtls_framework/outcome_analysis.py
```diff
@@ -112,11 +114,19 @@ def name_matches_pattern(name: str, str_or_re: IgnoreEntry) -> bool:
     else:
         return str_or_re == name
 
+def open_outcome_file(outcome_file: str) -> typing.TextIO:
+    if outcome_file.endswith('.gz'):
+        return gzip.open(outcome_file, 'rt', encoding='utf-8')
+    elif outcome_file.endswith('.xz'):
+        return lzma.open(outcome_file, 'rt', encoding='utf-8')
+    else:
+        return open(outcome_file, 'r', encoding='utf-8')
```
mpg, line 123: Question: why are we using `rt` for gzip/lzma open and just `r` here? (Not saying that's wrong, just curious.)

mpg (commented):
Looks pretty good to me, only one minor thing and a question.

mpg:
> Having moved files to the framework myself, I now realize that this move does not preserve file history.

For `collect_test_cases.py` and `outcome_analysis.py`, since they were created just before being moved, I'm not sure how useful it is to preserve their history. But for `check_test_cases.py` indeed that could be a nice thing.

eleuzi01:
> > Having moved files to the framework myself, I now realize that this move does not preserve file history.
> 
> For `collect_test_cases.py` and `outcome_analysis.py`, since they were created just before being moved, I'm not sure how useful it is to preserve their history. But for `check_test_cases.py` indeed that could be a nice thing.

Yes, I meant `check_test_cases.py` , doesn't seem that there is much history to be preserved though but maybe it's something to consider.

scripts/mbedtls_framework/outcome_analysis.py
gilles-peskine-arm: Because `t` is the default for the `open` builtin, but not for gzip and its imitators. And the existing code had `'r'` and I hadn't bothered to change it.

I'd actually considered writing this as
```
if …:
   open_function = …
else:
    open_function = open
open(outcome_file, 'r', encoding='utf-8')
```
but felt it was a bit too Haskelly for a Python script.

gilles-peskine-arm:
I agree that we should preserve the history. Even with the newly created files, we should keep the history around since their early history has content moved from other files. I'll try out [David's script](https://github.com/Mbed-TLS/mbedtls-docs/pull/145).

scripts/mbedtls_framework/outcome_analysis.py
mpg: That's probably how I would have written it, so I should probably check out Haskell next time I want to learn a new language ;)

gilles-peskine-arm:
I've addressed Manuel's comments. I intend to push a revised history that includes the history of the moved files and ends with identical content once I get David's script working.

scripts/mbedtls_framework/outcome_analysis.py
gilles-peskine-arm: Using an `open_function` variable doesn't work because the candidate functions have different types and (our current version of) mypy isn't smart enough to reconcile them without help. This isn't worth spending time figuring out if it can be helped.

mpg (commented):
Content LGTM at 53b6a463117b9c72403b0cb7a8f59de3e87c0669 - waiting for the version with identical content and improved history before I fully approve.

mpg:
I'm happy with the content, but the history seems to include some commits that are not relevant.

I mean, looking at [the](https://github.com/Mbed-TLS/mbedtls-framework/pull/55/commits/6901504ddb08d2e20e8a64f62d551106335154b3) [first](https://github.com/Mbed-TLS/mbedtls-framework/pull/55/commits/7a442c9941df1223508b4b68b9efd37b3acb95df) [three](https://github.com/Mbed-TLS/mbedtls-framework/pull/55/commits/bfbecf8b34dc362ffe194f5de2eec4031f41a557) they only touch `tests/ssl-opt.sh` and `library/ssl_xxx.c` so how are they relevant here?

Still approving as I don't think the extra commits are causing much harm, and I don't want this PR to get blocked while we sort out the script.

PR #58, opened 2024-10-10
PSA_WANT_xxx is only meaningful when PSA crypto is enabled

gilles-peskine-arm:
Fix a missing dependency in configuration symbol test enumeration: `PSA_WANT_xxx` is only meaningful when PSA crypto is enabled. This isn't needed in TF-PSA-Crypto where all crypto goes through PSA anyway, but it is relevant in 3.6 and harmless (as long as we keep `MBEDTLS_PSA_CRYPTO_CLIENT` at least as an internal symbol) in TF-PSA-Crypto.

* **development PR**: https://github.com/Mbed-TLS/mbedtls/pull/9593
* **3.6 PR**: https://github.com/Mbed-TLS/mbedtls/pull/9680

### DIFF
--- a/scripts/generate_config_tests.py
+++ b/scripts/generate_config_tests.py
@@ -80,7 +80,7 @@ def dependencies_of_setting(cfg: config_common.Config,
     be negated by prefixing them with '!'. This is the same syntax as a
     depends_on directive in test data.
     """
-    #pylint: disable=too-many-return-statements
+    #pylint: disable=too-many-branches,too-many-return-statements
     name = setting.name
     if name in SIMPLE_DEPENDENCIES:
         return SIMPLE_DEPENDENCIES[name]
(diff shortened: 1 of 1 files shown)


scripts/generate_config_tests.py
```diff
@@ -80,7 +80,7 @@ def dependencies_of_setting(cfg: config_common.Config,
     be negated by prefixing them with '!'. This is the same syntax as a
     depends_on directive in test data.
     """
-    #pylint: disable=too-many-return-statements
+    #pylint: disable=too-many-branches,too-many-return-statements
```
yanesca, line 83: Nitpick: instead of silencing the warning, we could move the TLS setting processing to a separate function.

yanesca (approved):
LGTM

paul-elliott-arm (approved):
LGTM

PR #61, opened 2024-10-24
Add a Python utility function to get the 3.6 feature macro

### LINKED ISSUE #51: Add a Python utility function to get the 3.6 feature macro for a given PSA algorithm and 3.6 domain
In the framework, when we generate test cases, we need to set dependency macros and dependency macros are not the same in 3.6 and 4.0. See #31 and #34 as examples of that. We prefer not to have an ad-hoc solution each time we encounter this issue in the Python scripts (like in #31 and #34). Thus the idea to have an utility function in test_case.py to help. The signature of this function has been discussed here https://github.com/Mbed-TLS/mbedtls-framework/pull/46#discussion_r1762973150. The goal of this task is to implement this utility function and use it to override the temporary solutions in #31 and #34.

eleuzi01:
Aims to resolve #51 

Development: https://github.com/Mbed-TLS/mbedtls/pull/9724
3.6: https://github.com/Mbed-TLS/mbedtls/pull/9725


scripts/mbedtls_framework/test_case.py
```diff
@@ -89,3 +111,28 @@ def write_data_file(filename: str,
             tc.write(out)
         out.write('\n# End of automatically generated file.\n')
     os.replace(tempfile, filename)
+
+def psa_or_3_6_feature_macro(psa_alg: str,
```
mpg, line 115: I think we need to better define what the `psa_alg` input is allowed to be. My thinking when I suggested this function was it would be similar to `psa_want_symbol`, ie it takes `PSA_ALG_xxx` or `PSA_KEY_TYPE_xxx` as input and returns the corresponding feature macro.

Here his function seems to accept both `PSA_WANT_xxx` and `PSA_ALG_xxx` as input. My first reaction is this looks like unnecessary complexity. However, is there a reason I'm missing?

Minor: since the function accepts not only algorithms but also key types, I don't think `psa_alg` is a great name. Perhaps `psa_name` (similarly to `psa_want_symbol()` that calls its argument `name`, but here we add `psa_` to distinguish from legacy). Or `psa_feature`?

Anyway, the docstring should clarify what kind of inputs are accepted.

scripts/mbedtls_framework/test_case.py
```diff
@@ -89,3 +111,28 @@ def write_data_file(filename: str,
...
+                             domain_3_6: str) -> str:
```
mpg, line 116: I'd prefer to use a [proper enum](https://docs.python.org/3/library/enum.html) for this parameter. Like, the class could be called `Domain_3_6` and then we have constants for `PSA`, `TLS1_3_ONLY`, `USE_PSA`, `LEGACY` so that the full name is like `Domain_3_6.USE_PSA` for example?

scripts/mbedtls_framework/test_case.py
```diff
@@ -10,12 +10,34 @@
...
+    "PSA_WANT_ALG_SHA_512" : "MBEDTLS_MD_CAN_SHA512",
+    "PSA_WANT_ALG_SHA3_224" : "MBEDTLS_MD_CAN_SHA3_224",
+    "PSA_WANT_ALG_SHA3_256" : "MBEDTLS_MD_CAN_SHA3_256",
+    "PSA_WANT_ALG_SHA3_384" : "MBEDTLS_MD_CAN_SHA3_384",
+    "PSA_WANT_ALG_SHA3_512" : "MBEDTLS_MD_CAN_SHA3_512"
+}
+
+PK_MACROS_3_6 = {
+    "PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY" : "MBEDTLS_PK_HAVE_ECC_KEYS",
+    "PSA_HAVE_ALG_SOME_ECDSA" : "MBEDTLS_PK_CAN_ECDSA_SOME",
```
mpg, line 33: I don't think we need the last three entries. `egrep 'MBEDTLS_PK_CAN_ECDSA_|SA_HAVE_ALG_ECDSA_|PSA_HAVE_ALG_SOME_ECDSA' **/*.py` finds no occurrence. And without them, we can have a simpler scheme where the input is always a PSA alg or key type (`PSA_ALG_xxx` or `PSA_KEY_TYPE_xxx` as opposed to `PSA_WANT_...` or `PSA_HAVE_...`).

(Also, I'm not sure is those are the correct translations.)

scripts/mbedtls_framework/test_case.py
```diff
@@ -89,3 +111,28 @@ def write_data_file(filename: str,
...
+            return psa_information.psa_want_symbol(psa_alg)
+
+    if psa_alg in HASHES_3_6:
+        return HASHES_3_6[psa_alg]
+    if psa_information.psa_want_symbol(psa_alg) in HASHES_3_6:
+        return HASHES_3_6[psa_information.psa_want_symbol(psa_alg)]
+    if psa_alg in PK_MACROS_3_6:
+        return PK_MACROS_3_6[psa_alg]
+
+    raise ValueError('Unable to determine dependency symbol for ' + psa_alg)
```
mpg, line 138: Should also print out the domain argument.

(Right now we don't seem to use it: hashes can always use `MD_CAN`, and `PK` is intrinsically in the `USE_PSA` domain.)

scripts/generate_pkcs7_tests.py
```diff
@@ -38,12 +38,7 @@ class TestData:
     Take in test_suite_pkcs7.data file.
     Allow for new tests to be added.
     """
-    #temporary solution to determine correct dependency macros between 3.6 and 4.0
-    #see issue #51 in mbedtls-framework
-    if build_tree.is_mbedtls_3_6():
-        mandatory_dep = "MBEDTLS_MD_CAN_SHA256"
-    else:
-        mandatory_dep = "PSA_WANT_ALG_SHA_256"
+    mandatory_dep = test_case.psa_or_3_6_feature_macro("PSA_ALG_SHA_256", "")
```
mpg, line 41: As a matter of principle, we should indicate the domain as being `USE_PSA` (PKCS7, like all X.509, is in this domain), even it it turns out it doesn't matter for hashes.

scripts/generate_config_tests.py
```diff
@@ -88,12 +87,8 @@ def dependencies_of_setting(cfg: config_common.Config,
...
             return 'MBEDTLS_CIPHER_C:MBEDTLS_CIPHER_MODE_CBC'
         if name.startswith('MBEDTLS_PK_PARSE_EC_'):
-            #temporary solution to determine correct dependency macros between 3.6 and 4.0
-            #see issue #51 in mbedtls-framework
-            if build_tree.is_mbedtls_3_6():
-                return 'MBEDTLS_PK_C:MBEDTLS_PK_HAVE_ECC_KEYS'
-            else:
-                return 'MBEDTLS_PK_C:PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY'
+            return 'MBEDTLS_PK_C:' + test_case.psa_or_3_6_feature_macro(
+                'PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY', '')
```
mpg, line 91: As a matter of principle, the domain should be `USE_PSA` for PK, even if the implementation will disregard that.

scripts/mbedtls_framework/test_case.py
```diff
@@ -89,3 +111,28 @@ def write_data_file(filename: str,
...
+            psa_alg[8:] in ['MD5', 'RIPEMD160', 'SHA_1', 'SHA_224', 'SHA_256',
+                            'SHA_384', 'SHA_512', 'SHA3_224',
+                            'SHA3_256', 'SHA3_384', 'SHA3_512']:
+            return psa_information.psa_want_symbol(psa_alg)
+
+    if psa_alg in HASHES_3_6:
+        return HASHES_3_6[psa_alg]
+    if psa_information.psa_want_symbol(psa_alg) in HASHES_3_6:
+        return HASHES_3_6[psa_information.psa_want_symbol(psa_alg)]
+    if psa_alg in PK_MACROS_3_6:
```
mpg, line 135: Perhaps add "and domain is `USE_PSA`" - otherwise it means the user did something wrong. (Then we'll fall through to the `raise ValueError` below.)

mpg (changes requested):
Solves the problem, but I think the new function's API can be improved and simplified.

scripts/mbedtls_framework/test_case.py
eleuzi01: I did this because while the issue description talked a lot about hashes, it didn't say anything about how to deal with PK symbols. And because I thought that we're also keeping `PSA_HAVE_ALG_SOME_ECDSA` and others that don't follow the same pattern as `PSA_ALG_xxx` or `PSA_KEY_TYPE_xxx`, I decided to cover both options for hashes knowing that one will probably be removed.

scripts/mbedtls_framework/test_case.py
eleuzi01: I included all PK symbols that I replaced (as part of my previous legacy symbol work) event though `PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY` is the only one used in the framework at the moment. Similarly, `PSA_ALG_SHA_256` is the only one used out of the hashes.

mpg: Ok, makes sense. I'm inclined to keep all the hashes, because handling all of them is no harder than handling just one, but drop the extra PK macros, as they might be trickier.

scripts/mbedtls_framework/test_case.py
mpg: Ok, makes complete sense. However, I happen to be on a personal crusade against complexity these days, so I'm still inclined to go for the simpler solution :)

eleuzi01: Totally agree, I'm still in the middle of reworking the PR. Just wanted to share why I did it 😄

scripts/mbedtls_framework/test_case.py
eleuzi01: The domains are still a bit confusing to me: from what I understand in 3.6 USE_PSA domain means we might use both PSA and legacy symbols depending on whether USE_PSA is enabled or not. But this function does not really reflect that with PK symbols.

mpg: Yes, things aren't very neat here. I think two things are going on:
1. The `psa_alg` and `domain` argument are not fully independent: all users of PK (and PK itself) are in the `USE_PSA` domain. Contrast with say AES: some users are in the legacy domain (for example PKCS5), some are in the `USE_PSA` domain (for example TLS), some in the PSA domain (PBKDF-AES-CMAC), so they're each going to use different macros for "is AES available".
2. Things in the `USE_PSA` domain, like PK, tend to define custom macros for "is XXX available" that take into account whether `USE_PSA` is enabled or not - so users don't have to be like "is `USE_PSA` on? then I'll use macro XXX, else I'll use macro YYY", then can just use the abstracted macro, like `MBEDTLS_PK_HAVE_ECC_KEYS` which expresses "PK has support for ECC keys (and you don't need to care whether it's based on PSA or legacy)".

I hope it makes sense. To be honest, this whole situation is really not an example of a clean architecture. From a purely engineering point of view, the right thing would have been to first deliver 4.0 where everything is based on PSA, and then work on driver-only builds. But we had to deliver driver-only earlier, so we did what we had to do, but the result isn't exactly pretty.

scripts/generate_config_tests.py
```diff
@@ -88,7 +88,7 @@ def dependencies_of_setting(cfg: config_common.Config,
             return 'MBEDTLS_CIPHER_C:MBEDTLS_CIPHER_MODE_CBC'
         if name.startswith('MBEDTLS_PK_PARSE_EC_'):
             return 'MBEDTLS_PK_C:' + test_case.psa_or_3_6_feature_macro(
-                'PSA_KEY_TYPE_ECC_PUBLIC_KEY', test_case.Domain_3_6.USE_PSA)
+                'PSA_KEY_TYPE_ECC_PUBLIC_KEY', test_case.Domain36.USE_PSA)
```
mpg, line 91: Ah, sorry for suggesting a name that wouldn't please pylint ^^

scripts/mbedtls_framework/test_case.py
```diff
@@ -89,3 +115,25 @@ def write_data_file(filename: str,
             tc.write(out)
         out.write('\n# End of automatically generated file.\n')
     os.replace(tempfile, filename)
+
+def psa_or_3_6_feature_macro(psa_name: str,
+                             domain_3_6: Domain36) -> str:
+    """Determine the dependency symbol for a given psa_alg based on
```
mpg, line 121: Minor: should rename to `psa_name`. Also, please clarify that the input is the name for the PSA mechanism itself, (ie `PSA_ALG`, `PSA_KEY_TYPE` etc), not the `PSA_WANT` macro.

scripts/mbedtls_framework/test_case.py
```diff
@@ -89,3 +115,25 @@ def write_data_file(filename: str,
...
+       the domain and Mbed TLS version. For more information about the domains,
+       and MBEDTLS_MD_CAN_ prefixed symbols, see transition-guards.md.
+       Currently works with hashes and PK symbols only.
```
mpg, line 124: Nit: and _some_ PK symbols.

mpg (changes requested):
Looks pretty good to me! Just some docstring tuning and we should be all good!

Harry-Ramsey (approved):
LGTM.

Harry-Ramsey (approved):
LGTM.

mpg:
@Harry-Ramsey Thanks for your review! As a general note, please remember to update labels after reviewing a PR. (You're not the only one who regularly forgets, we should all pay more attention.)

mpg:
@eleuzi01 Time to update the development and 3.6 PRs to point to the merge commit, and then we can merge them!

PR #62, opened 2024-10-25
Remove deprecated use of count as positional arg

paul-elliott-arm:
No intended change in functionality. Just removing the deprecated warning I am getting every time I generate test code.

Python 13 requires count to be passed as a kwarg rather than a positional arg because of potential confusion with flags, see https://github.com/python/cpython/issues/56166


gilles-peskine-arm (approved):
LGTM

scripts/generate_test_code.py
```diff
@@ -762,7 +762,7 @@ def escaped_split(inp_str, split_char):
         raise ValueError('Expected split character. Found string!')
     out = re.sub(r'(\\.)|' + split_char,
                  lambda m: m.group(1) or '\n', inp_str,
-                 len(inp_str)).split('\n')
+                 count = len(inp_str)).split('\n')
```
gilles-peskine-arm, line 765:
```suggestion
                 count=len(inp_str)).split('\n')
```

otherwise pylint will complain

gilles-peskine-arm: Until we have a CI in the framework repository, please at least run directly relevant tests. Here, at least `check-python-files.sh` (in a CI-like venv), and `make tests`.

PR #66, opened 2024-11-05
Support Unicode in .function files

### LINKED ISSUE #64: generate_test_code.py does not support Unicode in .function files
`generate_test_code.py` supports non-ASCII characters in `.data` files, but not portably in `.function` files. Sample backtrace (from https://github.com/Mbed-TLS/mbedtls/pull/9739 at https://github.com/Mbed-TLS/mbedtls/commit/db226b4eec73336eb61a4c9135bf2eed1c967896):
```
Traceback (most recent call last):
  File "../framework/scripts/generate_test_code.py", line 1294, in <module>
    main()
  File "../framework/scripts/generate_test_code.py", line 1289, in main
    c_file=out_c_file, out_data_file=out_data_file)
  File "../framework/scripts/generate_test_code.py", line 1217, in generate_code
    write_test_source_file(template_file, c_file, snippets)
  File "../framework/scripts/generate_test_code.py", line 1134, in write_test_source_file
    c_f.write(code)
UnicodeEncodeError: 'ascii' codec can't encode character '\xa7' in position 2012: ordinal not in range(128)
```

UTF-8 encoded characters work if the locale (`LC_CTYPE`) is based on UTF-8, or when using Python ≥3.7 in a non-Unicode locale. But on the CI, we try to use Python 3.6 and the locale is `C`, leading to the failure quoted above when I tried to use § in a `.function` file.

gilles-peskine-arm:
In the unit test code generator, use utf-8 when opening text files. Get rid of some cruft dating from the Python 2 days. Fix #64.

No prerequisites. Needed for https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/871.

### DIFF
--- a/scripts/audit-validity-dates.py
+++ b/scripts/audit-validity-dates.py
@@ -345,7 +345,7 @@ def parse_suite_data(filename: str) -> Iterator[Tuple[int, List[str]]]:
     :param data_f: file object of the data file.
     :return: Generator that yields test function argument list.
     """
-    with open(filename) as data_f:
+    with open(filename, encoding='utf-8') as data_f:
         for line_no, line in enumerate(data_f):
             line = line.strip()
             # Skip comments
--- a/scripts/audit-validity-dates.py
+++ b/scripts/audit-validity-dates.py
@@ -11,15 +11,14 @@ are valid throughout the desired validity period. The data are collected
 from framework/data_files/ and tests/suites/*.data files by default.
 """
 
-import os
-import re
-import typing
 import argparse
 import datetime
-import glob
-import logging
-import hashlib
 from enum import Enum
+import glob
+import hashlib
+import logging
+import os
+import re
 from typing import Callable, Dict, Iterator, List, Optional, Tuple
 
 # The script requires cryptography >= 35.0.0 which is only available
@@ -79,12 +78,11 @@ class X509Object(typing_util.Protocol):
 
 class AuditData:
     """Store data location, type and validity period of X.509 objects."""
-    #pylint: disable=too-few-public-methods
 
     def __init__(self, data_type: DataType, x509_obj: X509Object) -> None:
         self.data_type = data_type
         # the locations that the x509 object could be found
-        self.locations = [] # type: typing.List[str]
+        self.locations = [] # type: List[str]
         self.fill_validity_duration(x509_obj)
         self._obj = x509_obj
         encoding = cryptography.hazmat.primitives.serialization.Encoding.DER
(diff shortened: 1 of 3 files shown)


bjwtaylor (dismissed):

scripts/generate_test_code.py
```diff
@@ -228,57 +229,9 @@ class GeneratorInputError(Exception):
...
-        return self._line_no
-
-    @property
-    def name(self):
-        """
-        Property that indicates name of the file that is read.
-        """
-        return self._f.name
+def read_file(file_name: str) -> fileinput.FileInput:
+    return fileinput.input([file_name],
```
yanesca: fileinput.input() uses module-global state. That makes read_file() non-reentrant and potentially problematic if the generator is embedded, nested, or used concurrently. Constructing fileinput.FileInput(...) directly would avoid that:
```suggestion
    return fileinput.FileInput([file_name],
```

yanesca (dismissed):
There are some additional behavioural changes but most of them are an improvement and the rest is negligible in impact. I left a comment with a minor nit.

LGTM

yanesca:
Here are the bahavioural changes in case it helps with fixing the CI:
- the old wrapper stripped trailing whitespace from every input line; the replacement preserves it
- A filename equal to - now means standard input. The old wrapper opened a literal file named -
- Bare \r line endings are now recognized as separate lines. Previously only \n split lines.
- A final line without a newline no longer has one appended automatically.

yanesca (dismissed):
LGTM

scripts/audit-validity-dates.py
```diff
@@ -317,42 +345,42 @@ def parse_suite_data(data_f):
...
-
-        # Check parameters line
-        match = re.search(r'\A\w+(.*:)?\"', line)
-        if match:
-            # Read test vectors
-            parts = re.split(r'(?<!\\):', line)
-            parts = [x for x in parts if x]
-            args = parts[1:]
-            yield args
+    with open(filename) as data_f:
```
bjwtaylor: Would open(filename, encoding='utf-8') be better? As this will cause an error iterating # café or is this out of scope possible part of #64?

scripts/audit-validity-dates.py
```diff
@@ -317,42 +345,42 @@ def parse_suite_data(data_f):
...
+        for line_no, line in enumerate(data_f):
```
bjwtaylor: Should this be enumerate(data_f, 1) to start from line 1 not 0?

bjwtaylor (commented):
A couple of questions, let me know what you think?

bjwtaylor (approved):

yanesca (approved):
LGTM

PR #67, opened 2024-11-07
Add project and branch detection in shell

### LINKED ISSUE #39: Project and branch detection Python and shell utilities
When adapting Python and shell scripts to work for mbedtls 3.6, mbedtls 4.0 and/or TF-PSA-Crypto, we often need to detect if the script is called in mbedtls or TF-PSA-Crypto context, if it is called from the root of the project and in the case of mbedtls if the branch is 3.6 or 4.0. The goals of this task are:
1.  to complete/improve the Python utility functions in build_tree.py to do the above. Complete: no mbedlts 3.6/4.0 detection for the time being I think. Improve: to distinguish between mbedtls/TF-PSA-Crypto, find a project root, rather use the `scripts/project_name.txt` file introduced by https://github.com/Mbed-TLS/mbedtls/pull/8609 (to be added in tf-psa-crypto/scripts).
2. for shell scripts, extend along the lines of what has been done in 8609 to detect between mbedtls 3.6 and 4.0. We should probably create an utility script located in `framework/scripts` containing the functions `in_mbedtls_repo()` and `in_tf_psa_crypto_repo()` as defined in 8609 as well as an `is_3_6_branch()`/`is_4_x_branch()`. The utility scripts would be called from the tests scripts after having checked that we are at the root of mbedtls/TF-PSA-Crypto.

eleuzi01:
Resolves #39 

Development: https://github.com/Mbed-TLS/mbedtls/pull/9763
3.6: https://github.com/Mbed-TLS/mbedtls/pull/9764

Gatekeper note (mpg): CI can't pass here as we now depend on the `FRAMEWORK` variables defined by the associated PRs. So, we'll rely on CI status in 9763 and 9764. We also need to merge the 3 PRs at the same time in order to avoid breakage - so we won't do the usual pointer update in 9763 and 9764 after merging this one in order to reduce the window where things are in an inconsistent state. (We can update the pointer to the merge commit as follow-ups to 9763 and 9764.)

### DIFF
--- a/scripts/project_detection.sh
+++ b/scripts/project_detection.sh
@@ -1,0 +1,29 @@
+# project-detection.sh
+#
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+#
+# Purpose
+#
+# This script contains functions for shell scripts to
+# help detect which project (Mbed TLS, TF-PSA-Crypto)
+# or which Mbed TLS branch they are in.
+
+# Project detection
+read_project_name_file () {
+    PROJECT_NAME_FILE='../../scripts/project_name.txt'
+    if read -r PROJECT_NAME < "$PROJECT_NAME_FILE"; then :; else
+        echo "$PROJECT_NAME_FILE does not exist... Exiting..." >&2
+        exit 1
+    fi
+}
+
+in_mbedtls_repo () {
+    read_project_name_file
+    test "$PROJECT_NAME" = "Mbed TLS"
+}
+
+in_tf_psa_crypto_repo () {
+    read_project_name_file
+    test "$PROJECT_NAME" = "TF-PSA-Crypto"
+}
--- a/scripts/project_detection.sh
+++ b/scripts/project_detection.sh
@@ -27,3 +27,36 @@ in_tf_psa_crypto_repo () {
     read_project_name_file
     test "$PROJECT_NAME" = "TF-PSA-Crypto"
 }
+
+#Branch detection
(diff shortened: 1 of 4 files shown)


scripts/project_detection.sh
```diff
@@ -11,7 +11,16 @@
 
 # Project detection
 read_project_name_file () {
-    PROJECT_NAME_FILE='../../scripts/project_name.txt'
+    SCRIPT_DIR=$(pwd)
+
+    PROJECT_NAME_FILE="scripts/project_name.txt"
+
+    if echo "$SCRIPT_DIR" | grep -q "/framework/scripts" || echo "$SCRIPT_DIR" | grep -q "/tests/scripts"; then
```
mpg, line 19: Seems unnecessary if we assume the scripts are always invoked from the project's root (which as I said on the other PR, is a standard assumption - too bad it's often unwritten).

ronald-cron-arm:
@eleuzi01 As part of this PR or in a following one please use in_3_6_branch to improve what has been done in https://github.com/Mbed-TLS/mbedtls-framework/pull/76.

eleuzi01:
> @eleuzi01 As part of this PR or in a following one please use in_3_6_branch to improve what has been done in #76.

It might have to be done in a following PR because the CI is not passing when I do that. Either I'm doing something wrong or maybe there're still PRs that have to be merged before I'm able to remove this temporary solution?

mpg (approved):
LGTM, thanks!

davidhorstmann-arm (approved):
LGTM, thanks!

PR #76, opened 2024-11-15
Keep going if MBEDTLS_PSA_CRYPTO_CONFIG fails

ronald-cron-arm:
Companion PR of https://github.com/Mbed-TLS/mbedtls/pull/9771

### DIFF
--- a/scripts/all-helpers.sh
+++ b/scripts/all-helpers.sh
@@ -68,7 +68,10 @@ helper_libtestdriver1_adjust_config() {
     fi
 
     # Enable PSA-based config (necessary to use drivers)
-    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    # The configuration option has been removed for 4.0. While the project and
+    # branch detection shell in shell work is not completed, just ignore the
+    # failure to set MBEDTLS_PSA_CRYPTO_CONFIG.
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG || true
 
     # Dynamic secure element support is a deprecated feature and needs to be disabled here.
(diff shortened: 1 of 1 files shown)


scripts/all-helpers.sh
```diff
@@ -68,7 +68,10 @@ helper_libtestdriver1_adjust_config() {
     fi
 
     # Enable PSA-based config (necessary to use drivers)
-    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    # The configuration option has been removed for 4.0. While the project and
+    # branch detection shell in shell work is not completed, just ignore the
+    # failure to set MBEDTLS_PSA_CRYPTO_CONFIG.
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG > /dev/null 2>&1 || true
```
minosgalanakis, line 74: Should we use  `set +e` followed by a `set -e` so at least we don't silently hide the error?


~~~~~
set +e
scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG > /dev/null 2>&1 || true
set -e
~~~~~

This is not a strong ask, more so of a questoin

ronald-cron-arm: Changed to "scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG || true". That way we keep the log "A #define for the symbol MBEDTLS_PSA_CRYPTO_CONFIG was not found in include/mbedtls/mbedtls_config.h". And anyway as explained in the comment this is temporary until #9763 is completed.

minosgalanakis (approved):
LGTM

Harry-Ramsey (approved):
LGTM.

PR #77, opened 2024-11-18
Add out of source build functions for TF-PSA-Crypto

Harry-Ramsey:
This commit adds helper functions to build TF-PSA-Crypto out of source using CMake.

This commit is used for: https://github.com/Mbed-TLS/mbedtls/pull/9767

### DIFF
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -165,6 +165,7 @@ pre_initialize_variables () {
             PSA_CORE_PATH='tf-psa-crypto/core'
             BUILTIN_SRC_PATH='tf-psa-crypto/drivers/builtin/src'
             CONFIG_TEST_DRIVER_H='tf-psa-crypto/tests/configs/crypto_config_test_driver.h'
+            TF_PSA_CRYPTO_ROOT_DIR="$PWD/tf-psa-crypto"
         fi
         config_files="$CONFIG_H $CRYPTO_CONFIG_H $CONFIG_TEST_DRIVER_H"
     else
@@ -172,6 +173,7 @@ pre_initialize_variables () {
         PSA_CORE_PATH='core'
         BUILTIN_SRC_PATH='drivers/builtin/src'
         CONFIG_TEST_DRIVER_H='tests/configs/config_test_driver.h'
+        TF_PSA_CRYPTO_ROOT_DIR="$PWD"
 
         config_files="$CRYPTO_CONFIG_H $CONFIG_TEST_DRIVER_H"
     fi
@@ -937,6 +939,10 @@ run_component () {
         "${dd_cmd[@]}"
     fi
 
+    if in_tf_psa_crypto_repo; then
+        pre_create_tf_psa_crypto_out_of_source_directory
+    fi
+
     # Run the component in a subshell, with error trapping and output
     # redirection set up based on the relevant options.
     if [ $KEEP_GOING -eq 1 ]; then
@@ -971,8 +977,22 @@ run_component () {
     fi
 
     # Restore the build tree to a clean state.
+    if in_tf_psa_crypto_repo; then
+        cleanup_tf_psa_crypto_out_of_source_directory
+    fi
+
     cleanup
     unset current_component
+}
+
+pre_create_tf_psa_crypto_out_of_source_directory () {
+    rm -rf "$OUT_OF_SOURCE_DIR"
+    mkdir "$OUT_OF_SOURCE_DIR"
+    cd "$OUT_OF_SOURCE_DIR"
+}
+cleanup_tf_psa_crypto_out_of_source_directory () {
+    cd "$TF_PSA_CRYPTO_ROOT_DIR"
+    rm -rf "$OUT_OF_SOURCE_DIR"
 }
 
 ################################################################
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -1002,7 +1002,6 @@ run_component () {
 pre_create_tf_psa_crypto_out_of_source_directory () {
     rm -rf "$OUT_OF_SOURCE_DIR"
     mkdir "$OUT_OF_SOURCE_DIR"
-    cd "$OUT_OF_SOURCE_DIR"
 }
 
 ################################################################
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -948,11 +948,7 @@ run_component () {
         "${dd_cmd[@]}"
     fi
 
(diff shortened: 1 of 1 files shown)


scripts/all-core.sh
```diff
@@ -937,6 +941,15 @@ run_component () {
...
     fi
 
+    # Since building TF-PSA-Crypto is out of source, we cannot identify if we
+    # are in TF-PSA-Crypto repository. We cache the result, which is 0 for
+    # success.
+    in_tf_psa_crypto_repo
+    running_tf_psa_crypto_test="$?"
+    if [ "$running_tf_psa_crypto_test" -eq 0 ]; then
+        pre_create_tf_psa_crypto_out_of_source_directory
+    fi
```
gilles-peskine-arm: We run the script with `set -e` active. This means that if any command exit with a nonzero status, the script exits. Note that [this has a lot of limitations and pitfalls](https://mywiki.wooledge.org/BashFAQ/105), but those limitations aren't a concern here. Also, we run parts of the script with an EXIT trap which catches some here, but that too is not relevant at this particular point in the script.

When `all.sh` is running in the mbedtls repository, the command `in_tf_psa_crypto_repo` exits with the status 1. This causes the whole script to exit with 1.

Here the fix is to put the command directly in an if statement.

```suggestion
    if in_tf_psa_crypto_repo; then
        running_tf_psa_crypto_test=1
        pre_create_tf_psa_crypto_out_of_source_directory
    fi
```

Harry-Ramsey: Thank you for saving me that debugging headache, that makes a lot more sense to me now.

Harry-Ramsey:
I've ran a few various test cases for `MbedTLS 3.6` and cannot seem to understand why this error is occurring. I have tried one with all the changes commented out and no changes to the test cases there appeared.

ronald-cron-arm (changes requested):
One thing that can be removed now that all-core.sh does not cd automatically to the directory for the out of tree build, otherwise this looks good to me.

I think the component test_tf_psa_crypto_cmake_out_of_source of the CI against mbedtls:development fails because the component tries (as in version 9e4ac374e2 of TF-PSA-Crypto) to create the out of tree build directory while it already exists.

The CI against mbedtls:mbedtls-3.6 fails completely but that does not seem to be related to the content of this PR. Error line 232 in `framework/scripts/mbedtls_framework/config_common.py` but I do not understand the root cause.

scripts/all-core.sh
```diff
@@ -970,11 +989,21 @@ run_component () {
         fi
     fi
 
+    # Reset working directory to TF_PSA_Crypto as it is build out of source..
+    if [ $running_tf_psa_crypto_test -eq 1 ]; then
+        cd "$TF_PSA_CRYPTO_ROOT_DIR"
+    fi
+
```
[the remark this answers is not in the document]
Harry-Ramsey: I think this should be kept as components which build out of source or change directory can be reset to source directory without additional code required in those tests. Components which don't change directory will be unaffected by this line as they should already be in the source directory.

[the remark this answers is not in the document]
ronald-cron-arm: Please check with @mpg but I think components are run in a sub-shell thus if they change directory it does not matter to the calling shell.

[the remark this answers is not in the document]
gilles-peskine-arm: I haven't read this PR and haven't investigated whether this particular `cd` call is still needed. But I can confirm that components do run in a subshell, so if a component calls `cd` (or sets environment variables, and a few other things), this has no impact on subsequent components.

scripts/all-core.sh
```diff
@@ -165,13 +165,19 @@ pre_initialize_variables () {
             PSA_CORE_PATH='tf-psa-crypto/core'
             BUILTIN_SRC_PATH='tf-psa-crypto/drivers/builtin/src'
             CONFIG_TEST_DRIVER_H='tf-psa-crypto/tests/configs/crypto_config_test_driver.h'
+            MBEDTLS_ROOT_DIR="$PWD"
+            TF_PSA_CRYPTO_ROOT_DIR="$PWD/tf-psa-crypto"
+            MBEDTLS_FRAMEWORK_ROOT_DIR="$PWD/framework"
```
mpg, line 170: Question: we already have a variable `FRAMEWORK` defined in the top-level file (`all.sh`), why do we need that one in addition?

Harry-Ramsey: I believe that I added this before it was finalised where FRAMEWORK was going to be set as a temporary workaround. I can remove that now.

mpg (approved):
LGTM

ronald-cron-arm:
This looks good to me but unfortunately we are not able to validate the change against 3.6 with the framework CI. Thus we need to validate it with the CI of mbedtls:mbedtls-3.6. For this, we need in 9829 to just update the framework to the head of this PR: https://github.com/Mbed-TLS/mbedtls-framework/pull/77/commits/f3044f8d983e1fd66825bf94a4ce36e0500731b1. Please update 9829 that way, thanks.

Harry-Ramsey:
> This looks good to me but unfortunately we are not able to validate the change against 3.6 with the framework CI. Thus we need to validate it with the CI of mbedtls:mbedtls-3.6. For this, we need in 9829 to just update the framework to the head of this PR: [f3044f8](https://github.com/Mbed-TLS/mbedtls-framework/commit/f3044f8d983e1fd66825bf94a4ce36e0500731b1). Please update 9829 that way, thanks.

Resolved now, thanks.

ronald-cron-arm (approved):
LGTM and validated by the CI of https://github.com/Mbed-TLS/mbedtls/pull/9767 (mbedtls:development), https://github.com/Mbed-TLS/mbedtls/pull/9829 (mbedtls:mbedtls-3.6) and https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/121 (tf-psa-crypto:development).

ronald-cron-arm:
@Harry-Ramsey I've just merged the PR, please update https://github.com/Mbed-TLS/mbedtls/pull/9767, https://github.com/Mbed-TLS/mbedtls/pull/9829 and https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/121 with the commit id of the merge of this PR.

Harry-Ramsey:
I have donle #121 and #9829 but #9767 will have to wait until #121 is merged first.

ronald-cron-arm:
Yes that makes sense, we can leave the update of #9767 for when #121 is merged.

PR #56, opened 2024-10-07
Move TLS auxiliary test scripts to the framework

eleuzi01:
Resolves #54 

Development PR: https://github.com/Mbed-TLS/mbedtls/pull/9673

PR #80, opened 2024-11-20
Move scripts used by all-core.sh to the framework

### LINKED ISSUE #74: Move scripts used by `all-core.sh` to the framework
Move the following scripts to the framework:
```
scripts/output_env.sh
tests/scripts/quiet/*
```

Those are the scripts that are used by `all-core.sh`. As of writing they're identical between 3.6 and development, this task should be resolved by just moving them and adapting paths were needed.

eleuzi01:
Resolves #74 

Development PR: https://github.com/Mbed-TLS/mbedtls/pull/9788
3.6 PR: https://github.com/Mbed-TLS/mbedtls/pull/9789

### DIFF
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -266,8 +266,8 @@ setup_quiet_wrappers()
     # unless there is an error. This reduces logging overhead in the CI.
     #
     # Note that the cmake wrapper breaks unless we use an absolute path here.
-    if [[ -e ${PWD}/tests/scripts/quiet ]]; then
-        export PATH=${PWD}/tests/scripts/quiet:$PATH
+    if [[ -e ${PWD}/framework/scripts/quiet ]]; then
+        export PATH=${PWD}/framework/scripts/quiet:$PATH
     fi
 }
 
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -861,7 +861,7 @@ pre_check_tools () {
         *) set "$@" RUN_ARMCC=0;;
     esac
     # Use a path relative to the currently-sourced file.
-    "$@" "${BASH_SOURCE%/*}"/../../framework/scripts/output_env.sh
+    "$@" "${BASH_SOURCE%/*}"/output_env.sh
 }
 
 pre_generate_files() {
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -861,7 +861,7 @@ pre_check_tools () {
         *) set "$@" RUN_ARMCC=0;;
     esac
     # Use a path relative to the currently-sourced file.
-    "$@" "${BASH_SOURCE%/*}"/../../scripts/output_env.sh
+    "$@" "${BASH_SOURCE%/*}"/../../framework/scripts/output_env.sh
 }
 
 pre_generate_files() {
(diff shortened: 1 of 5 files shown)


eleuzi01:
When I did #56, I moved files from tests/scripts to framework/scripts. I don't really remember what was the reason behind that but I followed the same logic here. Could the reviewers let me know whether the files from tests/scripts should be under framework/test/scripts or just framework/scripts and I'll fix that if needed, thanks!

scripts/all-core.sh
```diff
@@ -861,7 +861,7 @@ pre_check_tools () {
         *) set "$@" RUN_ARMCC=0;;
     esac
     # Use a path relative to the currently-sourced file.
-    "$@" "${BASH_SOURCE%/*}"/../../scripts/output_env.sh
+    "$@" "${BASH_SOURCE%/*}"/../../framework/scripts/output_env.sh
```
mpg, line 864: This is correct but I think we can simplify this by removing `../../framework/scripts` now that the two files live in the same directory: `"${BASH_SOURCE%/*}"/output_env.sh` should do the trick.

mpg (dismissed):
LGTM. Just suggesting an optional minor improvement (I think/hope).

mpg:
The failures in the CI are expected and are fixed in the companion PRs which both have a green CI, so everything's OK.

We'll just need to merge the companion PRs quickly after merging this one.

eleuzi01:
> When I did #56, I moved files from tests/scripts to framework/scripts. I don't really remember what was the reason behind that but I followed the same logic here. Could the reviewers let me know whether the files from tests/scripts should be under framework/test/scripts or just framework/scripts and I'll fix that if needed, thanks!

@mpg this question got lost above, wdyt?

valeriosetti (dismissed):

mpg:
> When I did #56, I moved files from tests/scripts to framework/scripts. I don't really remember what was the reason behind that but I followed the same logic here. Could the reviewers let me know whether the files from tests/scripts should be under framework/test/scripts or just framework/scripts and I'll fix that if needed, thanks!
> @mpg this question got lost above, wdyt?

Ah indeed I had completely missed this question that ended up hidden in the long list of commits, thanks for raising it again!

To be honest I'm not sure, when I moved `all-core.sh` I did the same, mostly following what seemed to be the pattern from existing files. (Which is usually a good guess, but of course explicitly raising the question as you did here is better.)

I'm _assuming_ the reason is that the distinction between `scripts` and `tests/scripts` was never super useful in the first place, and makes things more complicated when it comes to importing python modules in one directory from modules/scripts in the other (currently we have a tool for doing it one way but not the other), so we decided to drop it while we're moving things.

Also, I guess since we're adding a new `framework` component to the path it's nice to remove the `tests` component at the same time to avoid making the tree too deep (though that's a relatively weak reason).

@ronald-cron-arm @gilles-peskine-arm Any thoughts you'd like to share?

mpg (approved):
LGTM, thanks!

valeriosetti (approved):

valeriosetti (approved):

mpg:
CI is green in both consuming PRs - which both point to the tip of this branch.

PR #81, opened 2024-11-20
Update for TF-PSA-Crypto `config.py`

gabor-mezei-arm:
Update `config_common.py` and test suite generation scripts for TF-PSA-Crypto `config.py`.

### DIFF
--- a/scripts/generate_config_tests.py
+++ b/scripts/generate_config_tests.py
@@ -5,6 +5,7 @@
 # Copyright The Mbed TLS Contributors
 # SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
 
+import inspect
 import re
 import sys
 from typing import Iterable, Iterator, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


ronald-cron-arm:
@gabor-mezei-arm if you rebase on top of main you should get a green CI. After updating the framework pointer accordingly in #9792 this should fix the framework conflict as well.

scripts/generate_config_tests.py
```diff
@@ -164,12 +165,20 @@ class ConfigTestGenerator(test_data_generation.TestGenerator):
...
 
     def __init__(self, settings):
-        self.mbedtls_config = config.MbedTLSConfig()
-        self.targets['test_suite_config.mbedtls_boolean'] = \
-            lambda: enumerate_boolean_setting_cases(self.mbedtls_config)
-        self.psa_config = config.CryptoConfig()
-        self.targets['test_suite_config.psa_boolean'] = \
-            lambda: enumerate_boolean_setting_cases(self.psa_config)
+        # pylint: disable=no-member
+        config_members = dict(inspect.getmembers(config))
```
minosgalanakis, line 169: Why is that needed? From what I [gather ](https://github.com/Mbed-TLS/mbedtls-framework/blob/a2c76945ca090f9dd099001d7c5158557f5a2036/scripts/mbedtls_framework/config_common.py#L79)MbedTLSConfig implements `__getitem__` and not `__dict__`. Why not keep it simple and use the `.items()` iterator instead?

We have already using metaclasses, but this method will strip the metaclass atributes, when they are not captures in a `__dict__` getter as per [documentation](https://docs.python.org/3/library/inspect.html#inspect.getmembers)

>>getmembers() will only return class attributes defined in the metaclass when the argument is a class and those attributes have been listed in the metaclass’ custom __dir__(). 

In short I do not understand why the added complexity?

minosgalanakis (commented):
Functionally it works, not sure why we need to add the inspect dependency, but could be missing something so a clarificaiton is needed.

scripts/generate_config_tests.py
gabor-mezei-arm: Both mbedtls and tf-psa-crypto have its own `config.py`. When it is loaded in `generate_config_tests.py` It is not known which one is loaded. With the `inspect` the module is checked which class it is implementing and doing stuff according to that. The complexity is needed for handling both `config.py`.

>Why is that needed? From what I [gather ](https://github.com/Mbed-TLS/mbedtls-framework/blob/a2c76945ca090f9dd099001d7c5158557f5a2036/scripts/mbedtls_framework/config_common.py#L79)MbedTLSConfig implements __getitem__ and not __dict__. Why not keep it simple and use the .items() iterator instead?

Here the `config` module is checked and not a `Config` class.

minosgalanakis (approved):
LGTM

ronald-cron-arm (approved):
LGTM

PR #83, opened 2024-11-21
Switch generate_psa_test.py to automatic dependencies for positive test cases

gilles-peskine-arm:
This is mostly refactoring around `generate_psa_tests.py` and supporting Python libraries. Create a class for test case generation that can automatically determine test case dependencies. Use this for `generate_psa_tests.py` for positive test cases.

Practical effect:

* Never-executed test cases are now easier to work with, because they're commented out and have the reason in a comment.
* Fix some test cases that were not properly detected as never-executed.

This is a step on the forward port of https://github.com/Mbed-TLS/mbedtls/pull/9025. Here, I handle the positive test cases of `generate_psa_tests.py`, which brings me to a natural cutoff point.

Follow-up: handling the negative test cases (`KeyTypeNotSupported`, `OpFail`).

## Review recommendations

The commits are broken down in a way to make their impact on the generated files easy to understand. Only three commits change the output, the others are pure refactoring:

* 119b3b9740b1a64d0ee8ca302a5dcd72b6d36d3d
* e3f9189103de9f14702fa8f8581457d179d8f0e4
* e1f38eb599fffb6b3ac14b087a5f38306d89279f

To validate my claims about the impact on the generated files, I use `mbedtls-trace-files.py` from https://github.com/Mbed-TLS/mbedtls-docs/pull/23 combined with the `diff-pairs` script attached here.
[diff-pairs.txt](https://github.com/user-attachments/files/18094084/diff-pairs.txt)

From the `framework` subdirectory with this pull request checked out, with https://github.com/Mbed-TLS/mbedtls/pull/9796 (3.6) checked out in the parent directory:
```
mbedtls-trace-files.py -b .. -f 1 $(git rev-parse ':/Create a framework')..HEAD $(./scripts/generate_psa_tests.py --list)
diff-pairs [0-9][0-9]-*/
ls -l [0-9][0-9]*.diff
```
Then for each `NN-SHA.diff` file, check that the content makes sense given the commit message for SHA.

With https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/ checked out in the parent, the following invocation of `mbedtls-trace-files.py` works:
```
mbedtls-trace-files.py --skip-make -r 'cd .. && framework/scripts/generate_psa_tests.py' -b .. -f 1 $(git rev-parse ':/Create a framework')..HEAD $(./scripts/generate_psa_tests.py --list)
diff-pairs [0-9][0-9]-*/
ls -l [0-9][0-9]*.diff
```

## PR checklist

- **crypto PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/122
- **development PR** https://github.com/Mbed-TLS/mbedtls/pull/9841
- **3.6 PR** https://github.com/Mbed-TLS/mbedtls/pull/9796

### DIFF
--- a/scripts/mbedtls_framework/psa_information.py
+++ b/scripts/mbedtls_framework/psa_information.py
@@ -8,7 +8,7 @@
 import os
 import re
 from collections import OrderedDict
-from typing import FrozenSet, List, Optional
+from typing import List, Optional
 
 from . import macro_collector
 
@@ -122,35 +122,6 @@ def generate_deps_from_description(
             dep_list += deps
 
     return dep_list
-
-# A temporary hack: at the time of writing, not all dependency symbols
-# are implemented yet. Skip test cases for which the dependency symbols are
-# not available. Once all dependency symbols are available, this hack must
-# be removed so that a bug in the dependency symbols properly leads to a test
-# failure.
-def read_implemented_dependencies(filename: str) -> FrozenSet[str]:
-    return frozenset(symbol
-                     for line in open(filename)
-                     for symbol in re.findall(r'\bPSA_WANT_\w+\b', line))
-_implemented_dependencies = None #type: Optional[FrozenSet[str]] #pylint: disable=invalid-name
-
-def find_dependencies_not_implemented(dependencies: List[str]) -> List[str]:
-    """List the dependencies that are not implemented."""
-    global _implemented_dependencies #pylint: disable=global-statement,invalid-name
-    if _implemented_dependencies is None:
-        # Temporary, while Mbed TLS does not just rely on the TF-PSA-Crypto
-        # build system to build its crypto library. When it does, the first
-        # case can just be removed.
-        if os.path.isdir('tf-psa-crypto'):
-            _implemented_dependencies = \
-                read_implemented_dependencies('tf-psa-crypto/include/psa/crypto_config.h')
-        else:
-            _implemented_dependencies = \
-                read_implemented_dependencies('include/psa/crypto_config.h')
-    return [dep
-            for dep in dependencies
-            if (dep.lstrip('!') not in _implemented_dependencies and
-                dep.lstrip('!').startswith('PSA_WANT'))]
 
 def tweak_key_pair_dependency(dep: str, usage: str):
     """
--- a/scripts/mbedtls_framework/psa_information.py
+++ b/scripts/mbedtls_framework/psa_information.py
@@ -23,8 +23,13 @@ class Information:
     def remove_unwanted_macros(
             constructors: macro_collector.PSAMacroEnumerator
     ) -> None:
-        # Mbed TLS does not support finite-field DSA.
+        """Remove constructors that should be exckuded from systematic testing."""
+        # Mbed TLS does not support finite-field DSA, but 3.6 defines DSA
+        # identifiers for historical reasons.
         # Don't attempt to generate any related test case.
+        # The corresponding test cases would be commented out anyway,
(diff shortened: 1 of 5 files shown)


davidhorstmann-arm:
Needs rebase atop #67

scripts/mbedtls_framework/psa_information.py
```diff
@@ -123,38 +139,6 @@ def generate_deps_from_description(
...
-            _implemented_dependencies = \
-                read_implemented_dependencies('include/psa/crypto_config.h')
-
-    if not all((dep.lstrip('!') in _implemented_dependencies or
-                not dep.lstrip('!').startswith('PSA_WANT'))
-               for dep in dependencies):
-        dependencies.append('DEPENDENCY_NOT_IMPLEMENTED_YET')
-
 def tweak_key_pair_dependency(dep: str, usage: str):
     """
```
gilles-peskine-arm, line 143: `psa_information.tweak_key_pair_dependency`, and thus `psa_test_case.TestCase.infer_dependencies`, silently suppresses dependencies on key pairs if a usage is specified and is not `BASIC` or `GENERATE` (for example `IMPORT` or `DERIVE`). Furthermore, if no usage is specified with `set_key_pair_usage()` and a key pair type is present, then `infer_dependencies` ends up generating the deprecated dependency symbol `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR`.

Right now, this ends up working, because we always call either `set_key_pair_usage('GENERATE')` or `set_key_pair_usage('BASIC')`. This could be considered in scope since `set_key_pair_usage()` is a new method and it isn't working correctly. This could also be considered out of scope since it doesn't break anything and the goal of this pull request is to forward-port some improvements, not to fix every problem we can find. I propose to consider it out-of-scope here, and handle it in the follow-up that handles negative test cases, for which the current behavior does introduce a bug.

reclaimer-chaplain: Agreed, let's consider it out of scope. (I am not reviewing, just commenting.)

yanesca: I agree, we should consider it out of scope. (Just commenting, not reviewing.)

gilles-peskine-arm:
With this branch as it is now (e1f38eb599fffb6b3ac14b087a5f38306d89279f), the CI [on 3.6](https://github.com/Mbed-TLS/mbedtls/pull/9796) passes but the CI [on
`mbedtls/development` post-split](https://github.com/Mbed-TLS/mbedtls/pull/9841) fails (it passed pre-split). This is due to unrelated changes in the framework that are required to get past the split.

I've switched the [mbedtls/development](https://github.com/Mbed-TLS/mbedtls/pull/9841) and [crypto](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/122) pull requests to a rebased framework branch, and I expect the CI to pass. We can either merge this PR as it is or add a merge commit with a recent `main` or rebase it on top of a recent `main`, to get the same content that (I expect) will pass the CI.

yanesca:
I think that if the PRs with the rebased framework branch pass, we should just merge this one as it is.

gabor-mezei-arm (approved):
Looks good to me. Ci is passing on all of the three using repos.

IMO the changes make the test generation code more readable.

waleed-elmelegy-arm (approved):
LGTM.

yanesca:
Merging as the crypto and TLS PRs passed CI (see https://github.com/Mbed-TLS/mbedtls-framework/pull/83#issuecomment-2551014364)

PR #88, opened 2024-12-03
Add a SBOM template in CycloneDX format

hughsie:
Improve supply chain security by including a SBOM file with substituted values.

This will be used to construct a composite platform SBOM.

See https://github.com/Mbed-TLS/mbedtls/pull/9777 for more information. But this can be merged independently.


gilles-peskine-arm (approved):
LGTM

PR #100, opened 2024-12-10
Move scripts/code_style.py to the framework

### LINKED ISSUE #70: Move `scripts/code_style.py` to the framework
This script has slightly diverged between development and 3.6. This task is to:
- reconcile the two branches;
- move the script to the framework;
- adapt the path everywhere the script is invoked.

Note: this task is only about mbedtls and the framework. There is a separate task for adapting this script to tf-psa-crypto: https://github.com/mbed-tls/tf-psa-crypto/issues/48 so that is out of scope here.

eleuzi01:
Resolves #70 

Development: https://github.com/Mbed-TLS/mbedtls/pull/9836
3.6: https://github.com/Mbed-TLS/mbedtls/pull/9837


minosgalanakis (approved):
Looks good. 

The failure is caused by `./tests/scripts/all.sh --seed 4 --keep-going check_python_files
`
Which is failing into the multibranch PR because of ` Duplicate module named "code_style" (also at "framework/scripts/code_style.py")`

This is a false positive and expected as we are moving things around.

PR #102, opened 2024-12-12
Add variable to identify ECC family

gabor-mezei-arm:
Add a new variable to `KeyType` to querry easily the ECC family name.


gabor-mezei-arm:
It is not needed anymore since all usage has been removed.

PR #115, opened 2025-01-08
TF-PSA-Crypto Doxygen Adaptations

Harry-Ramsey:
## Description
Adapt doxygen scripts to work for TF-PSA-Crypto.

## PR checklist

- [x] **crypto PR** Mbed-TLS/TF-PSA-Crypto: [#142](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/142)
- [x] **development PR** Mbed-TLS/mbedtls: not required.
- [x] **3.6 PR** Mbed-TLS/mbedtls: not required.

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push

### DIFF
--- a/scripts/apidoc_full.sh
+++ b/scripts/apidoc_full.sh
@@ -12,17 +12,48 @@
 
 set -eu
 
-CONFIG_H='include/mbedtls/mbedtls_config.h'
+. $(dirname "$0")/project_detection.sh
 
-if [ -r $CONFIG_H ]; then :; else
-    echo "$CONFIG_H not found" >&2
-    exit 1
+if in_mbedtls_repo; then
+    CONFIG_H='include/mbedtls/mbedtls_config.h'
+    if [ -r $CONFIG_H ]; then :; else
+        echo "$CONFIG_H not found" >&2
+    fi
+    if ! in_3_6_branch; then
+        CRYPTO_CONFIG_H='tf-psa-crypto/include/psa/crypto_config.h'
+    fi
 fi
 
-CONFIG_BAK=${CONFIG_H}.bak
-cp -p $CONFIG_H $CONFIG_BAK
+if in_tf_psa_crypto_repo; then
+    CRYPTO_CONFIG_H='include/psa/crypto_config.h'
+fi
 
-scripts/config.py realfull
-make apidoc
+if in_tf_psa_crypto_repo || (in_mbedtls_repo && ! in_3_6_branch); then
+    if [ -r $CRYPTO_CONFIG_H ]; then :; else
+        echo "$CRYPTO_CONFIG_H not found" >&2
+        exit 1
+    fi
+    CRYPTO_CONFIG_BAK=${CRYPTO_CONFIG_H}.bak
+    cp -p $CRYPTO_CONFIG_H $CRYPTO_CONFIG_BAK
+fi
 
-mv $CONFIG_BAK $CONFIG_H
+if in_mbedtls_repo; then
+    CONFIG_BAK=${CONFIG_H}.bak
+    cp -p $CONFIG_H $CONFIG_BAK
+    scripts/config.py realfull
+    make apidoc
+    mv $CONFIG_BAK $CONFIG_H
+elif in_tf_psa_crypto_repo; then
+    scripts/config.py realfull
+    TF_PSA_CRYPTO_ROOT_DIR=$PWD
+    rm -rf doxygen/build-apidoc-full
+    mkdir doxygen/build-apidoc-full
+    cd doxygen/build-apidoc-full
+    cmake -DCMAKE_BUILD_TYPE:String=Check -DGEN_FILES=ON $TF_PSA_CRYPTO_ROOT_DIR
+    make tfpsacrypto-apidoc
+    cd $TF_PSA_CRYPTO_ROOT_DIR
+fi
+
+if in_tf_psa_crypto_repo || (in_mbedtls_repo && ! in_3_6_branch); then
+    mv $CRYPTO_CONFIG_BAK $CRYPTO_CONFIG_H
+fi
--- a/scripts/doxygen.sh
+++ b/scripts/doxygen.sh
@@ -8,8 +8,10 @@
 # Abort on errors (and uninitialised variables)
 set -eu
 
-if [ -d library -a -d include -a -d tests ]; then :; else
-    echo "Must be run from Mbed TLS root" >&2
+. $(dirname "$0")/project_detection.sh
+
+if in_mbedtls_repo || in_tf_psa_crypto_repo; then :; else
+    echo "Must be run from Mbed TLS root or TF-PSA-Crypto root" >&2
     exit 1
 fi
 
@@ -28,5 +30,8 @@ if grep -E "(warning|error):" doc.filtered; then
     exit 1;
 fi
 
-make apidoc_clean
+if in_mbedtls_repo; then
+    make apidoc_clean
+fi
+
 rm -f doc.out doc.err doc.filtered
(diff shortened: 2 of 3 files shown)


scripts/apidoc_full.sh
```diff
@@ -22,7 +28,13 @@ fi
...
-scripts/config.py realfull
-make apidoc
-
-mv $CONFIG_BAK $CONFIG_H
+if in_mbedtls_repo; then
+    scripts/config.py realfull
+    make apidoc
+    mv $CONFIG_BAK $CONFIG_H
+elif in_tf_psa_crypto_repo; then
+    cd $OUT_OF_SOURCE_DIR
```
ronald-cron-arm, line 36:
```suggestion
    cd $OUT_OF_SOURCE_DIR
```
This is specific to all.sh and we want this script to work on its own. Here you can just build in tree:
`cmake -DCMAKE_BUILD_TYPE:String=Check .`.

scripts/doxygen.sh
```diff
@@ -28,5 +30,11 @@ if grep -E "(warning|error):" doc.filtered; then
     exit 1;
 fi
 
-make apidoc_clean
+if in_mbedtls_repo; then
+    make apidoc_clean
+elif in_tf_psa_crypto_repo; then
+    cd $TF_PSA_CRYPTO_ROOT_DIR
+    rm -rf $OUT_OF_SOURCE_DIR
```
ronald-cron-arm, line 37: ditto

ronald-cron-arm (changes requested):

ronald-cron-arm (changes requested):

scripts/apidoc_full.sh
```diff
@@ -12,17 +12,37 @@
...
 
-CONFIG_H='include/mbedtls/mbedtls_config.h'
+. $(dirname "$0")/project_detection.sh
 
-if [ -r $CONFIG_H ]; then :; else
-    echo "$CONFIG_H not found" >&2
+CRYPTO_CONFIG_H='include/psa/crypto_config.h'
+if in_mbedtls_repo; then
+    CONFIG_H='include/mbedtls/mbedtls_config.h'
+    CRYPTO_CONFIG_H='/tf-psa-crypto/include/psa/crypto_config.h'
```
ronald-cron-arm, line 20:
```suggestion
    CRYPTO_CONFIG_H='tf-psa-crypto/include/psa/crypto_config.h'
```

scripts/apidoc_full.sh
```diff
@@ -12,17 +12,37 @@
...
-scripts/config.py realfull
-make apidoc
+if in_mbedtls_repo; then
+    CONFIG_BAK=${CONFIG_H}.bak
+    cp -p $CONFIG_H $CONFIG_BAK
+    scripts/config.py realfull
+    make apidoc
+    mv $CONFIG_BAK $CONFIG_H
+elif in_tf_psa_crypto_repo; then
+    scripts/config.py full
```
ronald-cron-arm, line 43:
```suggestion
    scripts/config.py realfull
```
and you need to add support for realfull in config.py in TF-PSA-Crypto as done in config.py in mbedtls. This should fix the "documentation for unknown define MBEDTLS_xyz found." warnings.

ronald-cron-arm (changes requested):

scripts/apidoc_full.sh
```diff
@@ -12,17 +12,37 @@
...
-make apidoc
+if in_mbedtls_repo; then
+    CONFIG_BAK=${CONFIG_H}.bak
+    cp -p $CONFIG_H $CONFIG_BAK
+    scripts/config.py realfull
+    make apidoc
+    mv $CONFIG_BAK $CONFIG_H
+elif in_tf_psa_crypto_repo; then
+    scripts/config.py realfull
+    cmake -DCMAKE_BUILD_TYPE:String=Check -DGEN_FILES=ON .
```
ronald-cron-arm, line 44: Having played with it, it would be better to build out of tree in a temporary directory. Otherwise when we run `doxygen.sh` locally the source tree ends up with all the CMake artefacts. `mktemp` is it seems the way to create a temporary directory in bash.

ronald-cron-arm (changes requested):

valeriosetti (dismissed):

scripts/apidoc_full.sh
```diff
@@ -12,17 +12,40 @@
...
 
-CONFIG_H='include/mbedtls/mbedtls_config.h'
+. $(dirname "$0")/project_detection.sh
 
-if [ -r $CONFIG_H ]; then :; else
-    echo "$CONFIG_H not found" >&2
+CRYPTO_CONFIG_H='include/psa/crypto_config.h'
+if in_mbedtls_repo; then
+    CONFIG_H='include/mbedtls/mbedtls_config.h'
+    CRYPTO_CONFIG_H='tf-psa-crypto/include/psa/crypto_config.h'
```
ronald-cron-arm, line 20: This does not work in 3.6 case.

ronald-cron-arm (changes requested):
To fix the CI test_full_cmake_clang and test_psa_compliance jobs you need to rebase on top of main.

scripts/apidoc_full.sh
gilles-peskine-arm: `tmp=$(mktemp -d)` creates a temporary directory in a user-configurable location (defaulting to `/tmp`) and stores the location in `$tmp`. `tmp=$(TMPDIR=/path/to/parent mktemp -d)` creates a temporary directory under `/path/to/parent`. Avoid other options to `mktemp` for portability to older Linux and \*BSD.

But do we really need a temporary directory name? We want to keep the Doxygen build products around, so why not use a fixed directory name like `doxygen/build-apidoc-full`?

ronald-cron-arm: The documentation is currently generated at the root of the source tree in the `apidoc`  directory thus we do not need to keep the build tree. A fixed build directory in the source tree would not really improve things regarding not modifying the build tree thus not sure it would be an improvement.

gilles-peskine-arm: A fixed build directory is slightly easier to program, and is easier to inspect and clean if an error occurs.

ronald-cron-arm: okay the main thing I wanted to avoid here was to build in tree. I am fine with a build directory `doxygen/build-apidoc-full`.

scripts/apidoc_full.sh
```diff
@@ -12,17 +12,53 @@
...
 
-CONFIG_BAK=${CONFIG_H}.bak
-cp -p $CONFIG_H $CONFIG_BAK
+if in_mbedtls_repo; then
+    CRYPTO_CONFIG_H='tf-psa-crypto/include/psa/crypto_config.h'
+    if [ -r $CRYPTO_CONFIG_H ]; then :; else
+        echo "$CRYPTO_CONFIG_H not found" >&2
+        exit 1
+    fi
+fi
```
ronald-cron-arm, line 30:
```
if in_mbedtls_repo; then                                                        
    CONFIG_H='include/mbedtls/mbedtls_config.h'                                 
    if [ -r $CONFIG_H ]; then :; else                                           
        echo "$CONFIG_H not found" >&2                                          
    fi                                                                          
    if ! in_3_6_branch; then                                                    
        CRYPTO_CONFIG_H='tf-psa-crypto/include/psa/crypto_config.h'             
    fi                                                                          
fi
```

ronald-cron-arm (changes requested):
A few things to fix, otherwise we do not plan to have mbedtls PRs for this thus we need a green CI and for that we need to rebase against latest `main`.

ronald-cron-arm (approved):
LGTM, thanks.

valeriosetti (approved):

PR #116, opened 2025-01-09
Move pkgconfig.sh to the framework

### LINKED ISSUE #101: Move `pkgconfig.sh` to the framework
Move `tests/scripts/pkgconfig.sh` to the framework.

This script has diverged between development and 3.6. This task is to:

- reconcile the two branches;
- move the script to the framework;
- adapt the path everywhere the script is invoked.

Note: this task is only about mbedtls and the framework. This script is supposed to be already adapted to tf-psa-crypto; we just need to take care not to undo that while reconciling the two mbedtls branches - this will likely mean aligning on the development version in most cases, but please consider for each case if that is indeed the right thing to do.

valeriosetti:
## Description

Resolves #101 

## PR checklist

- [ ] **crypto PR** n/a
- [ ] **development PR** Mbed-TLS/mbedtls#9888
- [ ] **3.6 PR** Mbed-TLS/mbedtls#9889

**Note to self**: the file was moved with:
```
../mbedtls-docs/tools/bin/mbedtls-move-to-framework \
--from . --to ../mbedtls-framework \
--src-branch issue101-development \
--dst-branch issue101-framework \
--path tests/scripts/pkgconfig.sh:scripts/pkgconfig.sh
```

### DIFF
--- a/scripts/pkgconfig.sh
+++ b/scripts/pkgconfig.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+#
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+#
+# Purpose
+#
+# Test pkgconfig files.
+#
+# For each of the build pkg-config files, .pc files, check that
+# they validate and do some basic sanity testing on the output,
+# i.e. that the strings are non-empty.
+#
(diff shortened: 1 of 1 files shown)


ronald-cron-arm (approved):
LGTM

Harry-Ramsey (approved):
LGTM

PR #117, opened 2025-01-09
Adapt check_names.py for independent TF-PSA-Crypto testing

Harry-Ramsey:
This commit separates CodeParser into three classes. CodeParser a base class containing methods for parsing .c and .h files as well as retrieving symbols from compiled libraries.

As well as two subclasses of CodeParser: TF-PSA-Crypto_CodeParser and MBEDTLS_CodeParser, which are responsible for parsing the TF-PSA-Crypto and Mbed TLS libraries.

## PR checklist

- [ ] **crypto PR** Mbed-TLS: [#143](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/143)
- [ ] **development PR** Mbed-TLS: [#9915](https://github.com/Mbed-TLS/mbedtls/pull/9915)

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push

### DIFF
--- a/scripts/check_names.py
+++ b/scripts/check_names.py
@@ -212,7 +212,8 @@ class CodeParser():
     """
     def __init__(self, log):
         self.log = log
-        build_tree.check_repo_path()
+        if not build_tree.looks_like_root(os.getcwd()):
+            raise Exception("This script must be run from Mbed TLS or TF-PSA-Crypto root")
 
         # Memo for storing "glob expression": set(filepaths)
         self.files = {}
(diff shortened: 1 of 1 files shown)


Harry-Ramsey:
The last remaining error is to do with pylint. `__parse` has too many arguments however for simplicity I believe this should be left as it is.

scripts/check_names.py
```diff
@@ -1016,8 +1164,21 @@ def main():
...
+            tf_psa_crypto_code_parser = TFPSACryptoCodeParser(log)
+            tf_psa_crypto_parse_result = tf_psa_crypto_code_parser.comprehensive_parse()
+            mbedtls_code_parser = MBEDTLSCodeParser(log)
+            mbedtls_parse_result = mbedtls_code_parser.comprehensive_parse()
+            # Combine parse results together for NameChecker
+            parse_result = {}
+            for key in tf_psa_crypto_parse_result:
+                parse_result[key] = tf_psa_crypto_parse_result[key] + mbedtls_parse_result[key]
+        else:
+            raise Exception("This script must be run from Mbed TLS or TF-PSA-Crypto root")
```
valeriosetti, line 1181: Since CodeParser class already checks if the script is being run from the proper folder in its `_init_()` function, what about removing the `else` branch and simplifying the rest as follows:
```suggestion
        # Mbed TLS uses TF-PSA-Crypto, so we parse TF-PSA-Crypto in both cases
        tf_psa_crypto_code_parser = TFPSACryptoCodeParser(log)
        parse_result = tf_psa_crypto_code_parser.comprehensive_parse()

        if build_tree.looks_like_mbedtls_root(os.getcwd()):
            mbedtls_code_parser = MBEDTLSCodeParser(log)
            mbedtls_parse_result = mbedtls_code_parser.comprehensive_parse()
            # Combine parse results together for NameChecker
            parse_result = {}
            for key in tf_psa_crypto_parse_result:
                parse_result[key] = tf_psa_crypto_parse_result[key] + mbedtls_parse_result[key]
```
?

valeriosetti (changes requested):

scripts/check_names.py
Harry-Ramsey: Thanks for the suggestion. I have refactored the code :)

valeriosetti (dismissed):
LGTM :)

scripts/check_names.py
```diff
@@ -753,15 +783,18 @@ def parse_symbols(self):
             )
             my_environment = os.environ.copy()
             my_environment["CFLAGS"] = "-fno-asynchronous-unwind-tables"
```
valeriosetti, line 785: This is partially pre-existing (because this line was already there, but there's also a copied one below now) and also minor improvement, so it's up to you if it's worth to be done or not.
Since I wasn't aware of this flag and it took me some time to figure out what it does, the same might happen also to other people in the future, so I think a short comment on what this flag does would be nice. Wdyt?

scripts/check_names.py
```diff
@@ -727,7 +623,141 @@ def parse_identifiers(self, include, exclude=None):
...
+
+    def comprehensive_parse(self):
+        """
+        (Must be defined as a class method)
+        Comprehensive ("default") function to call each parsing function and
+        retrieve various elements of the code, together with the source location.
+
+        Returns a dict of parsed item key to the corresponding List of Matches.
+        """
+        raise NotImplementedError("comprehension_parse must be implemented by a code parser")
```
ronald-cron-arm, line 643: What about to just remove them from CodeParser?

scripts/check_names.py
```diff
@@ -727,7 +623,141 @@ def parse_identifiers(self, include, exclude=None):
...
+                    symbols.append(symbol.group("symbol"))
+                else:
+                    self.log.error(line)
+        return symbols
+
+class TFPSACryptoCodeParser(CodeParser):
+    """
+    Class for retrieving files and parsing TF-PSA-Crypto code. This can be used
+    independently of the checks that NameChecker performs, for example for
+    list_internal_identifiers.py.
```
ronald-cron-arm, line 683:
```suggestion
    Class for retrieving files and parsing TF-PSA-Crypto code. This can be used
    independently of the checks that NameChecker performs.
```
The reference to list_internal_identifiers.py seems quite irrelevant here.

scripts/check_names.py
```diff
@@ -771,64 +804,180 @@ def parse_symbols(self):
...
             )
 
         return symbols
 
-    def parse_symbols_from_nm(self, object_files):
+class MBEDTLSCodeParser(CodeParser):
+    """
+    Class for retrieving files and parsing Mbed TLS code. This can be used
+    independently of the checks that NameChecker performs, for example for
+    list_internal_identifiers.py.
```
ronald-cron-arm, line 831:
```suggestion
    Class for retrieving files and parsing Mbed TLS code. This can be used
    independently of the checks that NameChecker performs.
```
ditto

scripts/check_names.py
```diff
@@ -1016,15 +1165,21 @@ def main():
...
-        code_parser = CodeParser(log)
-        parse_result = code_parser.comprehensive_parse()
+        tf_psa_crypto_code_parser = TFPSACryptoCodeParser(log)
+        parse_result = tf_psa_crypto_code_parser.comprehensive_parse()
+        if build_tree.looks_like_mbedtls_root(os.getcwd()):
+            mbedtls_code_parser = MBEDTLSCodeParser(log)
+            mbedtls_parse_result = mbedtls_code_parser.comprehensive_parse()
+            # Combine parse results together for NameChecker
+            for key in mbedtls_parse_result:
+                parse_result[key] += mbedtls_parse_result[key]
```
ronald-cron-arm, line 1175: I would put this logic rather in MBEDTLSCodeParser. MBEDTLSCodeParser parses all mbedtls code, TF-PSA-Crypto code included in 4.x case. Internally it uses TFPSACryptoCodeParser to parse the TF-PSA-Crypto code. And here we have just something like (not tested):
```
if build_tree.looks_like_mbedtls_root(os.getcwd()):
    code_parser = MBEDTLSCodeParser(log)
    parse_result = mbedtls_code_parser.comprehensive_parse()
elsif build_tree.looks_like_tf_psa_crypto_root(os.getcwd()):
    code_parser = TFPSACryptoCodeParser(log)
    parse_result = tf_psa_crypto_code_parser.comprehensive_parse()
```

scripts/check_names.py
Harry-Ramsey: I think these should stay to retain compatibilty. These are abstract methods that all code parsers, present and future must implement to be used interchangabily.

This should hopefully enforce a more rigid code parser test script.

scripts/check_names.py
```diff
@@ -727,7 +623,141 @@ def parse_identifiers(self, include, exclude=None):
...
+    independently of the checks that NameChecker performs, for example for
+    list_internal_identifiers.py.
+    """
+
+    def __init__(self, log):
+        super().__init__(log)
+        if build_tree.looks_like_tf_psa_crypto_root(os.getcwd()):
+            self.source_dir = os.getcwd()
+        else:
+            self.source_dir = os.path.join(os.getcwd(), "tf-psa-crypto")
```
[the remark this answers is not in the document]
Harry-Ramsey: This check is just here because we can be inside Mbed TLS and run check_names.py so we need to locate TF-PSA-Crypto source. In TF-PSA-Crypto we just use the current working directory for source path.

ronald-cron-arm:
CI is not happy.

ronald-cron-arm (changes requested):

ronald-cron-arm (changes requested):
This looks almost good to me, the comment about "pylint" should be address to appease the CI, otherwise minor improvement suggestions.

ronald-cron-arm (approved):
LGTM, thanks.

scripts/check_names.py
Harry-Ramsey: I think just to prevent issues down the line, we might want to define this flag before each build. It is slightly more verbose (even though duplicated) and prevents any issues with environment changes between the two function calls.

I know that in [C](https://www.evanjones.ca/setenv-is-not-thread-safe.html) at least the environment variable is not reliable between threads and potentially subprocesses :)

valeriosetti (approved):

PR #130, opened 2025-01-28
tests: fix overlength buffer

billatarm:
The buffer was defined as 32 characters, but the string was 33 characters when including the NULL byte.

This led to build errors on Fedora:
/builddir/build/BUILD/mbedtls3.6-3.6.2-build/mbedtls-3.6.2/tests/src/psa_exercise_key.c:186:36: error: initializer-string for array of ‘unsigned char’ is too long [-Werror=unterminated-string-initialization]
  186 |     unsigned char ciphertext[32] = "(wabblewebblewibblewobblewubble)";

Fix this by leaving out the size portion of the array definition becuase C89 allows this.

## Description

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2340827

## PR checklist

- [ ] **crypto PR** Mbed-TLS/TF-PSA-Crypto#
- [ ] **development PR** Mbed-TLS/mbedtls#

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push

### DIFF
--- a/tests/src/psa_exercise_key.c
+++ b/tests/src/psa_exercise_key.c
@@ -184,7 +184,7 @@ static int exercise_cipher_key(mbedtls_svc_key_id_t key,
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
     psa_key_type_t key_type;
     const unsigned char plaintext[16] = "Hello, world...";
-    unsigned char ciphertext[32] = "(wabblewebblewibblewobblewubble)";
+    unsigned char ciphertext[] = "(wabblewebblewibblewobblewubble)";
     size_t ciphertext_length = sizeof(ciphertext);
     unsigned char decrypted[sizeof(ciphertext)];
     size_t part_length;


billatarm:
There are a whole pile of these, will re-open later.

tests/src/psa_exercise_key.c
```diff
@@ -184,7 +184,7 @@ static int exercise_cipher_key(mbedtls_svc_key_id_t key,
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
     psa_key_type_t key_type;
     const unsigned char plaintext[16] = "Hello, world...";
-    unsigned char ciphertext[32] = "(wabblewebblewibblewobblewubble)";
+    unsigned char ciphertext[] = "(wabblewebblewibblewobblewubble)";
```
gilles-peskine-arm, line 187: When the algorithm is one of `PSA_ALG_ECB_NO_PADDING` or `PSA_ALG_CBC_NO_PADDING`, `sizeof(plaintext)` needs to be a multiple of the block size, which is 8 or 16 depending on the key type. When the given usage includes DECRYPT but not ENCRYPT, this constraint also applies to `sizeof(ciphertext)`. (When the usage includes both DECRYPT and ENCRYPT, the length of the encrpytion result is used instead of `sizeof(ciphertext)`. For completeness, for decrypt-only usage, the algorithm `PSA_ALG_CBC_PKCS7` is also problematic.)

It happens that we currently never call this function with a usage that has DECRYPT without ENCRYPT. (Which surprised me a little, and suggests a test gap, but not a very important one.) So the tests are still passing with this patch, but the function is no longer doing its job in the problematic case.

I'd expect something similar to come up in several other places where `-Wunterminated-string-initialization` complains.

Note that in crypto code, we don't have null-terminated strings. When we write an `unsigned char` array in test code, we treat it as a byte array and there's no particular reason why the last byte would be 0. Here the code is correct and is what was intended.

We do intend a clean build with GCC 15, but since it isn't released yet, silencing its new warning isn't a high priority for us at the moment.

Please note that there's another change in GCC 15 that's much more important: [with the union initialization weakening, it breaks our code](https://github.com/Mbed-TLS/mbedtls/issues/9814). For the time being, you need to compile with `-fzero-init-padding-bits=unions`. We intend to fix this, because it's a bug in our code. But due to the complexity of a complete fix (including builds with drivers, which isn't a concern for Linux distributions), this may take a little time, so it may or may not be ready in time for the next minor release of Mbed TLS. I would expect the unit tests (`make test`) to fail if there is a bug due to union initialization (as long as you don't plug in secure element drivers).

billatarm: I have a few patches for `-Wunterminated-string-initialization` changes now, I can submit them if you'd like? Also, I see the other bug #9814, I also have a few patches there, but I can throw that compiler flag in on the Fedora package for now.

Issue #87, opened 2024-12-02
Move misc. "easy" script to the framework


mpg:
Move the following files to the framework:
```
tests/scripts/psa_collect_statuses.py
tests/scripts/run-metatests.sh
tests/scripts/run_demos.py
tests/scripts/test_zeroize.gdb
```

These are the scripts that are:
- currently identical between 3.6 and development;
- not covered by a previous issue.

To be discussed: should we also move `programs/test/zeroize.c` and `programs/test/metatest.c`?

Note: as usual for this EPIC, the work is only to move the files to the framework repo and adapt the paths everywhere they're used in mbedtls 3.6 and development. If further changes are needed to adapt to tf-psa-crypto, they're out of scope and will be covered by separate tasks.

gilles-peskine-arm:
> `tests/scripts/psa_collect_statuses.py`

I wrote that a while ago to solve a short-term problem (coverage of error cases and comparing that with what the specification draft says, to identify missing cases in the specification and missing coverage in our tests). I got it merged because it seemed like it could be a useful way to measure this aspect of test coverage, but I don't think we've made any use of it since then. So we could remove it and the corresponding code, if that's easier than moving it to the framework.

Today I would use the PSA wrapper generator framework for this, and I even have code to do that (also logging parameters, not just return values) in https://github.com/Mbed-TLS/mbedtls/pull/8559 (I used that partly to validate the PSA wrapper generator framework and partly because I needed to log the coverage of parameters).

> `programs/test/zeroize.c`

It's very tightly coupled with `test_zeroize.gdb` so I think both should be in the same repository. And I can't think of a reason why they should be different in different branches, so it makes sense for both to be in the framework.

> `programs/test/metatest.c?`

In principle this could have branch-specific code. But right now I can't think of anything branch-specific that I'd want to add even if I had plenty of spare time. It's mostly about checking that we're catching the kinds of failure we expect to catch from the test framework as well as from environment features such as sanitizers. So I think it belongs together with `tests/src`, in the framework.

mpg:
My main question about programs is are we ready to have C executable programs (as opposed to libraries / object files) in the framework, in terms of how our build system(s) work? Cc @ronald-cron-arm 

Other than that, I'm also inclined to have them in the framework. But if that's not easy, we might want to post-pone it (or at least split it to a separate task).

ronald-cron-arm:
[...]

PR #131, opened 2025-01-29
Move programs from mbedtls to framework

Harry-Ramsey:
## Description
Move program files to MbedTLS-Framework.

## PR checklist

- [ ] **crypto PR** Mbed-TLS/TF-PSA-Crypto#
- [ ] **development PR** Mbed-TLS/mbedtls#

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push


ronald-cron-arm (changes requested):
We have lost `test_zeroize.gdb` here and https://github.com/Mbed-TLS/mbedtls/pull/9939 and https://github.com/Mbed-TLS/mbedtls/pull/9939 are still expecting it to be in the framework. Their CI is green because they point to https://github.com/Mbed-TLS/mbedtls-framework/commit/fb9a8b383a9506449c5a51e0d52267fa169fa9f2, penultimate version of this PR.

Otherwise as you have to change this PR, `dlopen_demo.sh` and `test_zeroize.gdb` are all related to a test program thus they would fit better in `tests/programs`. No need to rework the "move" commit for that, you can just add a new commit that move the scripts from `scripts` to `tests/programs` in the framework.

tests/programs/test_zeroize.gdb
```diff
@@ -0,0 +1,64 @@
...
+# Note: This test requires that the test program is compiled with -g3.
+
+set confirm off
+
+file $zeroize_exe
+
+search GDB_BREAK_HERE
+break $_
+
+set args $framework_dir/tests/programs/zeroize.c
```
ronald-cron-arm, line 41: In the previous commit (fb9a8b383a9506449c5a51e0d52267fa169fa9f2) that was working I think it was:
```
set args ./framework/tests/programs/zeroize.c
```
Where do you expect `framework_dir` to be set up?

Harry-Ramsey: This is an argument that will be passed into the script using `-ex "set \$framework_dir=$FRAMEWORK`. This would enable the script to work for MbedTLS and TF-PSA-Crypto unless the zeroize test is being removed from MbedTLS in which case this can be reverted.

tests/programs/dlopen_demo.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
```
gilles-peskine-arm, line 1: Note that we have a program `run_demos.py`, invoked by `all.sh` components, which runs all the scripts called `*_demo.sh`. Please make sure to update this script so that it looks in all `framework/tests/programs`.

On a related note, [`run_demos.py` should move to the framework](https://github.com/Mbed-TLS/mbedtls-framework/issues/87). It may be easier to do this before changing it.

gilles-peskine-arm: Reviewers beware: there's no automatic mechanism that would let us know if we stop running the demo programs on the CI. They aren't even recorded in the outcome file. So please check the CI logs.

ronald-cron-arm: Good point, thanks. However, currently,  dlopen_demo.sh is actually never run through run_demos.py as there is no component where shared libraries are built and run_demos.py is run. dlopen_demo.sh is run directly in `test_make_shared` and `test_cmake_shared`.

ronald-cron-arm: This PR is already quite big, I prefer do that rather as part of https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/127. I've added a note there to not forget about it.

tests/programs/dlopen_demo.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Run the shared library dynamic loading demo program.
+# This is only expected to work when Mbed TLS is built as a shared library.
+
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+
+. "${0%/*}../../scripts/demo_common.sh"
```
ronald-cron-arm, line 9:
```suggestion
. "${0%/*}/../../scripts/demo_common.sh"
```
With that fix the CI of #9939 should be green. Otherwise the framework pointer in #9988 does not seem to be correct.

ronald-cron-arm (changes requested):

scripts/demo_common.sh
```diff
@@ -0,0 +1,137 @@
...
+  shift
+  echo "+ ! $*"
+  not "$@"
+}
+
+## config_has SYMBOL...
+## Succeeds if the library configuration has all SYMBOLs set.
+config_has () {
+  for x in "$@"; do
+    "$programs_dir/test/query_compile_time_config" "$x"
```
valeriosetti, line 87: Not sure this is OK. On line 43 we're setting
`programs_dir="$root_dir/programs"`
which means this line would be:
`"$root_dir/programs/test/query_compile_time_config" "$x"`
but `query_compile_time_config` is being moved in this PR into the framework repo. Am I wrong?

Harry-Ramsey: I believe that is the executable. The files have been moved but the executable is still built in the programs directory.

valeriosetti: Oh right, I missed that. Question though: is there any reason for which we move source files in the framework, but we keep the generated program file in the Mbed TLS main repo?

ronald-cron-arm (changes requested):
I've done a complete review now. Just some comments that need to be updated.

scripts/demo_common.sh
Harry-Ramsey: From my understanding it is because we do not want any build objects/generated files in the framework as this is shared between MbedTLS and TF-PSA crypto. There may be additional reasons others can clarify :)

ronald-cron-arm: That way we have all the test programs in the same place. The ones built from only branch specific code like `benchmark`, the ones built with a mix of branch specific code and framework code like `query_compile_time_config` and the ones built only with framework code.

ronald-cron-arm (approved):
LGTM, thanks.

valeriosetti (approved):

PR #132, opened 2025-01-29
[Framework] Move test_psa_*.py scripts to the framework

valeriosetti:
## Description

This is the `framework` part of https://github.com/Mbed-TLS/mbedtls/pull/9940

## PR checklist

- [x] **crypto PR** not required
- [x] **development PR** Mbed-TLS/mbedtls#9940
- [x] **3.6 PR** Mbed-TLS/mbedtls#9941


Harry-Ramsey (dismissed):
LGTM

scripts/test_psa_compliance.py
```diff
@@ -0,0 +1,172 @@
...
+        # Temporary while the PSA compliance test suite is still run as part
+        # of Mbed TLS testing. When it is not the case anymore, the second case
+        # can be removed.
+        if in_tf_psa_crypto_repo:
+            extra_includes = ';{}/drivers/builtin/include'.format(root_dir)
+        elif os.path.isdir(os.path.join(root_dir, 'tf-psa-crypto')):
+            extra_includes = ';{}/tf-psa-crypto/include'.format(root_dir) + \
+                             (';{}/tf-psa-crypto/drivers/builtin/include'.format(root_dir))
+        else:
+            extra_includes = ''
```
ronald-cron-arm, line 90:
```suggestion
        elif is_mbedtls_3_6:
            extra_includes = ''
        else:
            os.path.isdir(os.path.join(root_dir, 'tf-psa-crypto')):
            extra_includes = ';{}/tf-psa-crypto/include'.format(root_dir) + \
                             (';{}/tf-psa-crypto/drivers/builtin/include'.format(root_dir))           
```
with is_mbedtls_3_6 set where in_tf_psa_crypto_repo is.
In framework CI, when testing against 3.6 there is an empty tf-psa-crypto directory.

ronald-cron-arm (changes requested):
This looks almost good to me but one pre-existing condition that I'd like to improve.

scripts/test_psa_compliance.py
valeriosetti: > In framework CI, when testing against 3.6 there is an empty tf-psa-crypto directory.

I wasn't aware of this "trick", thanks for letting me know. I'll fix it immediately

scripts/test_psa_compliance.py
```diff
@@ -83,7 +83,9 @@ def main(library_build_dir: str):
         # can be removed.
         if in_tf_psa_crypto_repo:
             extra_includes = ';{}/drivers/builtin/include'.format(root_dir)
-        elif os.path.isdir(os.path.join(root_dir, 'tf-psa-crypto')):
+        elif build_tree.is_mbedtls_3_6():
```
ronald-cron-arm, line 86: At this point we have moved out of the mbedtls tree thus it probably does not work. Rather as we have an `in_tf_psa_crypto_repo` locla variable we should have an `is_mbedtls_3_6` one or probably even better move the  setting of `extra_includes` at the beginning of the function.

ronald-cron-arm (changes requested):

scripts/test_psa_compliance.py
valeriosetti: I'm going to implement this change as you suggested, but I have a question/curiosity on this
> At this point we have moved out of the mbedtls tree thus it probably does not work

Since both https://github.com/Mbed-TLS/mbedtls/pull/9940 and https://github.com/Mbed-TLS/mbedtls/pull/9941 have a green CI, how can we tell that this change is not working?

ronald-cron-arm: Good point. Actually the current directory is a descendant of the mbedtls root directory and there is a `guess_project_root` thus it was actually correct. A bit fragile probably thus the new way is quite better to me anyway.

ronald-cron-arm (changes requested):
Just one comment in check-python-files.sh that would be nicer to update:
```
#     tests/scripts/test_psa_constant_names.py:34: error: Cannot find implementation or library stub for module named 'mbedtls_framework'
```

ronald-cron-arm (approved):
LGTM

Harry-Ramsey (approved):

ronald-cron-arm:
Validated by the CI of https://github.com/Mbed-TLS/mbedtls/pull/9940 (dev) and https://github.com/Mbed-TLS/mbedtls/pull/9941 (3.6).

PR #135, opened 2025-02-05
Fix union initialization in PSA operations for GCC 15 (test helpers)

gilles-peskine-arm:
Test helpers and additional checks for the fix to https://github.com/Mbed-TLS/mbedtls/issues/9814.

Continues https://github.com/Mbed-TLS/mbedtls-framework/pull/136. Will not pass the CI without some bug fixes from https://github.com/Mbed-TLS/mbedtls/pull/9955.

## PR checklist

- [ ] **TF-PSA-Crypto PR** TODO
- [ ] **development PR** TODO
- [x] **3.6 PR** provided

PR #136, opened 2025-02-06
Fix union initialization in PSA operations for GCC 15: new test helpers

gilles-peskine-arm:
New test helpers that will be used for the fix of https://github.com/Mbed-TLS/mbedtls/issues/9814.

This just adds new functions that can be consumed by new test code (https://github.com/Mbed-TLS/mbedtls/pull/9955 — I'm planning some history rewrite, but no code changes except to handle CI and review feedback). There will be a follow-up in the framework that adds new checks (https://github.com/Mbed-TLS/mbedtls-framework/pull/135/commits/bd33fb4d457de54b6ed1b555db3e3408dc557682) that will only pass once the library is fixed. (At the moment, the new checks fail after `abort()` for some operations, because in abort functions, we consistently clean up core metadata and sensitive data in operation objects, but we sometimes leave driver-specific metadata behind, which makes the setup-time is-all-zero check fail when an operation object is reused after abort/finish.)

## PR checklist

- [ ] **TF-PSA-Crypto PR** TODO (just to update the submodule)
- [ ] **development PR** TODO (just to update the submodule)
- [ ] **3.6 PR** TODO (just to update the submodule)


gilles-peskine-arm:
Looking further, there's no good reason to split the framework part. On its own, the whole framework part doesn't break the CI. So everything can go into https://github.com/Mbed-TLS/mbedtls-framework/pull/135, we don't need a separate “preliminaries” pull request in the framework.

tests/include/test/drivers/test_driver_common.h
```diff
@@ -0,0 +1,11 @@
+/* Common definitions used by test drivers. */
+/*  Copyright The Mbed TLS Contributors
+ *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+ */
+
+#ifndef PSA_CRYPTO_TEST_DRIVERS_TEST_DRIVER_COMMON_H
+#define PSA_CRYPTO_TEST_DRIVERS_TEST_DRIVER_COMMON_H
+
+#include "mbedtls/build_info.h"
```
mpg, line 9: Note: currently this is already included by all users before deciding whether to include this file. But it doesn't hurt now and might be helpful later, so no objection.

tests/include/test/psa_crypto_helpers.h
```diff
@@ -145,6 +145,35 @@ const char *mbedtls_test_helper_is_psa_leaking(void);
     while (0)
 
 
+/** Initializer that doesn't set the embedded union to zero.
```
mpg, line 148: Note: I grepped `include/psa' for `union {`, found 15 occurrences, and verify that all of them chain to a structure that's in the list below - which has 3 extras entries that don't have a union yet. Some structures have more than one union, directly or indirectly, here the larger number of matches for `union {` then structures below.

mpg (approved):
LGTM, thanks!

bjwtaylor (approved):

mpg:
@gilles-peskine-arm This is fully approved and has a green CI (Open), so I'm inclined to merge it. I see the description has "TODO" for the crypto/TLS/3.6 PRs but I don't think we need to wait for that in order to merge this one, considering it has a green CI already?

gilles-peskine-arm:
This just needs submodule updates in the consuming branches, then it can start being consumed. Merging it would be good (and thanks for the reviews). The CI result is three months old though, so I'll just merge `main` and rerun the CI to make sure this hasn't bitrotted.

PR #141, opened 2025-02-25
[framework] Add test_tf_psa_crypto_cmake_shared to components-build-system.sh

valeriosetti:
## Description

Part of https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/138

## PR checklist

- [x] **TF-PSA-Crypto PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/181
- [x] **development PR** https://github.com/Mbed-TLS/mbedtls/pull/10008
- [x] **3.6 PR** not required

### DIFF
--- a/scripts/demo_common.sh
+++ b/scripts/demo_common.sh
@@ -16,18 +16,43 @@
 
 set -e -u
 
-## $root_dir is the root directory of the Mbed TLS source tree.
+DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG=${DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG:-1}
+
+need_query_compile_time_config () {
+  if [ $DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG -eq 1 ]; then
+    return 0;
+  else
+    return 1;
+  fi
+}
+
+## At the end of the while loop below $root_dir will point to the root directory
+## of the Mbed TLS or TF-PSA-Crypto source tree.
 root_dir="${0%/*}"
-# Find a nice path to the root directory, avoiding unnecessary "../".
-# The code supports demo scripts nested up to 4 levels deep.
-# The code works no matter where the demo script is relative to the current
-# directory, even if it is called with a relative path.
-n=4 # limit the search depth
-while ! [ -d "$root_dir/programs" ] || ! [ -d "$root_dir/library" ]; do
+## Find a nice path to the root directory, avoiding unnecessary "../".
+##
+## The code supports demo scripts nested up to 4 levels deep.
+##
+## The code works no matter where the demo script is relative to the current
+## directory, even if it is called with a relative path.
+n=5
+while true; do
+  # If we went up too many folders, then give up and return a failure.
   if [ $n -eq 0 ]; then
     echo >&2 "This doesn't seem to be an Mbed TLS source tree."
     exit 125
   fi
+  # If we reached the Mbed TLS root folder then we're done.
+  if is_mbedtls_root "$root_dir"; then
+    break;
+  fi
+  # If we reached the TF-PSA-Crypto root folder and the script that sourced
+  # this file does not need query_compile_time_config (which is only available
+  # in Mbed TLS repo) then we're done.
+  if is_tf_psa_crypto_root "$root_dir" && ! need_query_compile_time_config; then
+    break;
+  fi
+
   n=$((n - 1))
   case $root_dir in
     .) root_dir="..";;
@@ -37,10 +62,6 @@ while ! [ -d "$root_dir/programs" ] || ! [ -d "$root_dir/library" ]; do
     *) root_dir=".";;
   esac
 done
-
-## $programs_dir is the directory containing the sample programs.
-# Assume an in-tree build.
-programs_dir="$root_dir/programs"
 
 ## msg LINE...
 ## msg <TEXT_ORIGIN
@@ -80,9 +101,20 @@ run_bad () {
   not "$@"
 }
 
+## $programs_dir is the directory containing the sample programs.
+## Assume an in-tree build.
+programs_dir="$root_dir/programs"
+
 ## config_has SYMBOL...
 ## Succeeds if the library configuration has all SYMBOLs set.
+##
+## Note: "query_compile_time_config" is only available when in Mbed TLS project,
+## so "config_has" is not available in TF-PSA-Crypto one. If this function is
+## called from the latter we fail immediately.
 config_has () {
+  if ! is_mbedtls_root "$root_dir"; then
+    return 1;
+  fi
   for x in "$@"; do
     "$programs_dir/test/query_compile_time_config" "$x"
   done
--- a/scripts/project_detection.sh
+++ b/scripts/project_detection.sh
@@ -9,25 +9,79 @@
 # help detect which project (Mbed TLS, TF-PSA-Crypto)
 # or which Mbed TLS branch they are in.
 
-# Project detection
+# Project detection.
+#
+# Both Mbed TLS and TF-PSA-Cryto repos have a "scripts/project_name.txt" file
+# which contains the name of the project. They are used in scripts to know in
+# which project/folder we're in.
+# This function accepts 2 parameters:
+# - $1: boolean value which defines the behavior in case
+#       "scripts/project_name.txt" is not found:
+#       - 1: exit with error message
+#       - 0: simply return an error
+# - $2: mandatory value which defined the root folder where to look for
+#       "scripts/project_name.txt".
 read_project_name_file () {
-    SCRIPT_DIR=$(pwd)
+    EXIT_IF_NOT_FOUND=$1
+    ROOT_PATH=$2
 
     PROJECT_NAME_FILE="scripts/project_name.txt"
 
-    if read -r PROJECT_NAME < "$PROJECT_NAME_FILE"; then :; else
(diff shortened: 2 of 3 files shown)


Harry-Ramsey (dismissed):

tests/programs/dlopen_demo.sh
```diff
@@ -6,17 +6,29 @@
 # Copyright The Mbed TLS Contributors
 # SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
 
-. "${0%/*}/../../scripts/demo_common.sh"
+SCRIPT_DIR=$(dirname "$0")
+. "${SCRIPT_DIR}/../../scripts/demo_common.sh"
```
ronald-cron-arm, line 10: In a pure TF-PSA-Crypto context (not TF-PSA-Crypto inside Mbed TLS), when I run `./tests/scripts/all.sh test_tf_psa_crypto_cmake_shared` with https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/181, I get `This doesn't seem to be an Mbed TLS source tree.` when running `$FRAMEWORK/tests/programs/dlopen_demo.sh`. I think this is because `demo_common.sh` needs to be adapted to TF-PSA-Crypto.

ronald-cron-arm (changes requested):

Harry-Ramsey (dismissed):

scripts/demo_common.sh
```diff
@@ -16,14 +16,32 @@
...
+##
+## The code works no matter where the demo script is relative to the current
+## directory, even if it is called with a relative path.
+##
+## This search is based on the fact that both Mbed TLS and TF-PSA-Crypto has
+## a file in "<root>/scripts/project_name.txt" which contains the name of the
+## project. We take advantage of this file to check if we reached the right path.
 n=4 # limit the search depth
-while ! [ -d "$root_dir/programs" ] || ! [ -d "$root_dir/library" ]; do
+while ! [ -f "$root_dir/scripts/project_name.txt" ]; do
```
ronald-cron-arm, line 44: This breaks `tf-psa-crypto/programs/psa/key_ladder_demo.sh` and `tf-psa-crypto/programs/psa/key_ladder_demo.sh` and thus `tests/scripts/run_demos.py` as `root_dir` then corresponds to `tf-psa-crypto` and there is no `query_compile_time_config` in TF-PSA-Crypto currently. What about to first search for mbedtls root and if we do not find it, search for TF-PSA-Crypto root?

scripts/demo_common.sh
```diff
@@ -128,7 +159,7 @@ trap 'cleanup; trap - HUP; kill -HUP $$' HUP
 trap 'cleanup; trap - INT; kill -INT $$' INT
 trap 'cleanup; trap - TERM; kill -TERM $$' TERM
 
-if config_has MBEDTLS_ENTROPY_NV_SEED; then
+if [ $IS_MBEDTLS_ROOT -eq 1 ] && config_has MBEDTLS_ENTROPY_NV_SEED; then
```
ronald-cron-arm, line 162: I would rather in `config_has` return systematically "do not have" in TF-PSA-Crypto case.

ronald-cron-arm (changes requested):

valeriosetti:
@ronald-cron-arm  I think I've addressed all the problem with testing from various root folders (Mbed TLS vs TF-PSA-Crypto). The PRs series should be ready for review again

Harry-Ramsey (approved):

ronald-cron-arm (changes requested):
A few comments/suggestions.

scripts/demo_common.sh
```diff
@@ -16,18 +16,67 @@
...
+
+# Check if $1 is the root of the project specified by $2.
+check_root_type () {
+  read -r PROJECT_NAME < "$1/$PROJECT_NAME_FILE"
+  if [ "$PROJECT_NAME" = "$2" ]; then
+    return 0;
+  else
+    return 1;
+  fi
+}
```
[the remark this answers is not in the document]
valeriosetti: Yes, I can move these functions to `project_detection.sh`, but I found then easier to parse that script from `*_demo.sh` files rather than `demo_common.sh`. The reason is that from the latter I know the cwd, whereas for the latter I don´t.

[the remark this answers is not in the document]
ronald-cron-arm: I think `*_demo.sh` files include `demo_common.sh` thus they have the same cwd, but I may be wrong there.

scripts/project_detection.sh
```diff
@@ -9,28 +9,62 @@
...
 in_tf_psa_crypto_repo () {
     read_project_name_file
     test "$PROJECT_NAME" = "TF-PSA-Crypto"
 }
 
+# Check is $1 is either an Mbed TLS or TF-PSA-Crypto root folder.
+#
+# Differently from in_mbedtls_repo() and in_tf_psa_crypto_repo() above, this
+# can be run on any folder without causing the script to exit.
+is_valid_root () {
```
ronald-cron-arm, line 46: Instead of `is_valid_root` and `check_root_type` as new utility functions having `is_mbedtls_root` and `is_tf_psa_crypto_root` would be better I think. Same as the related `in_mbedtls/tf_psa_crypto_root` but they take one path argument. Then `read_project_name_file` would handle zero or one path argument. In case of zero argument it takes as currently the current working directory. With these new utility functions we should be able to get rid of the IS_MBEDTLS_ROOT environment variable.

scripts/demo_common.sh
```diff
@@ -16,18 +16,48 @@
...
+## Find a nice path to the root directory, avoiding unnecessary "../".
+##
+## The code supports demo scripts nested up to 4 levels deep.
+##
+## The code works no matter where the demo script is relative to the current
+## directory, even if it is called with a relative path.
+##
+## This search is based on the fact that both Mbed TLS and TF-PSA-Crypto has
+## a file in "<root>/scripts/project_name.txt" which contains the name of the
+## project. We take advantage of this file to check if we reached the right path.
```
ronald-cron-arm, line 41:
```suggestion
```
this is rather `is_mbedtls/tf_psa_crypto_root` documentation.

ronald-cron-arm (changes requested):

scripts/project_detection.sh
valeriosetti: Currently `read_project_name_file` fails if `PROJECT_NAME_FILE` is not found. This is something I avoided in `is_valid_root` so that I can call that function and prevent the calling script to exit while looking for the proper root.
I can however modify `read_project_name_file` and `in_mbedtls_repo`/`in_tf_psa_crypto_repo` in order to reduce the number of functions.

tests/programs/dlopen_demo.sh
```diff
@@ -6,17 +6,31 @@
...
+. "${SCRIPT_DIR}/../../scripts/project_detection.sh"
+. "${SCRIPT_DIR}/../../scripts/demo_common.sh"
 
 msg "Test the dynamic loading of libmbed*"
 
-program="$programs_dir/test/dlopen"
-library_dir="$root_dir/library"
+# Once demo_common.sh is sourced we're granted to be either in Mbed TLS or
+# TF-PSA-Crypto root folders and IS_MBEDTLS_ROOT variable can be used to
+# discriminate between the two.
```
ronald-cron-arm, line 19: IS_MBEDTLS_ROOT has been removed and this is a bit misleading I'd say. We are granted to be in Mbed TLS or TF-PSA-Crypto tree but not to be at the root of the tree (if I am not missing something) thus we should keep using the `programs_dir` and `root_dir` variables, not `pwd`.

ronald-cron-arm (changes requested):
This looks almost good to me. I have only quite minor comments.

scripts/project_detection.sh
```diff
@@ -9,28 +9,89 @@
...
+is_mbedtls_root() {
+    if ! read_project_name_file false $1 ; then
+        return 1
+    fi
+
+    if [ "$PROJECT_NAME" = "Mbed TLS" ]; then
+        return 0
+    else
+        return 1
+    fi
```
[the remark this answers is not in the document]
valeriosetti: I thought about this as well, but then I thought that I need to add an extra `return $?` below the `test` command. However checking the Bash documentation it seems that the return value of the function is the return of the last executed command, so this makes your suggestion way more appealing than mine :)

tests/programs/dlopen_demo.sh
```diff
@@ -6,17 +6,35 @@
...
+# - $root_dir points to the root path of Mbed TLS or TF-PSA-Crypto;
+# - $programs_dir points to "$root_dir/programs" folder.
+if is_mbedtls_root $root_dir; then
+    msg "Running in Mbed TLS repo"
+    program="$programs_dir/test/dlopen"
+    library_dir="$root_dir/library"
+else
+    msg "Running in TF-PSA-Crypto repo"
+    build_dir="."
+    if [ -d "$root_dir/out_of_source_build" ]; then
```
ronald-cron-arm, line 27: `out_of_source_build` is all.sh specific thus we cannot really do that here. I think I understand why you do that though and I would rather suggest to just build in tree in `component_test_tf_psa_crypto_shared`.

ronald-cron-arm (changes requested):
Minor comments but one that needs to be addressed I think.

ronald-cron-arm (approved):
LGTM, thanks.

Harry-Ramsey (approved):

PR #140, opened 2025-02-21
[Framework] md: allow dispatch to PSA whenever CRYPTO_CLIENT is enabled

valeriosetti:
## Description

This is the companion PR for https://github.com/Mbed-TLS/mbedtls/pull/9562

## Important note for reviewers

This PR is a _clone_ of https://github.com/Mbed-TLS/mbedtls-framework/pull/128 and the reason to clone it is that in #128 I somehow managed to push my working branch to `mbedtls-framework` repo instead of my own. Apparently this was allowed back then, but then restrictions were applied so that I cannot update this branch anymore. Therefore I pushed my working branch to my repo and opened a new PR.

## PR checklist

Please add the numbers (or links) of the associated pull requests for consuming branches. You can omit branches where this pull request is not needed.

- [ ] crypto PR Mbed-TLS/TF-PSA-Crypto#
- [ ] development PR Mbed-TLS/mbedtls#
- [ ] 3.6 PR https://github.com/Mbed-TLS/mbedtls/pull/9562

PR #142, opened 2025-03-01
Generate TLS handshake defragmentation tests

gilles-peskine-arm:
Python script to generate TLS handshake defragmentation test cases for `ssl-opt.sh`. This has the same coverage as the ones currently in the [defragmentation feature branch (development version)](https://github.com/Mbed-TLS/mbedtls/tree/features/tls-defragmentation/development).

- [x] **changelog** provided | not required because: test only
- [x] **development PR** https://github.com/Mbed-TLS/mbedtls/pull/10021
- [x] **TF-PSA-Crypto PR** not required because: TLS only
- [x] **framework PR** here
- [x] **3.6 PR** https://github.com/Mbed-TLS/mbedtls/pull/10029
- [x] **2.28 PR** not required because: new feature
- **tests**  provided


scripts/generate_tls_handshake_tests.py
```diff
@@ -13,6 +13,137 @@
...
+
+    our_args = ''
+    their_args = ''
+
+    if length is None:
+        description = 'no fragmentation, for reference'
+    else:
+        description = 'len=' + str(length)
+    if version is not None:
+        description += ', TLS 1.' + str(version.value)
```
mpg, line 51: Not a big fan of accessing `version.value` here and below. Can we have helper methods instead? For example:
- `__str__` -> `TLS 1.x`
- `openssl_opt` -> `-tls1_x`
- etc.

scripts/generate_tls_handshake_tests.py
```diff
@@ -13,6 +13,137 @@
...
+
+    # To guarantee that the handhake messages are large enough and need to be
+    # split into fragments, the tests require certificate authentication.
+    # The party in control of the fragmentation operations is OpenSSL and
+    # will always use server5.crt (548 Bytes).
+    if length is not None and \
+       length >= TLS_HANDSHAKE_FRAGMENT_MIN_LENGTH:
+        tc.requirements.append('requires_certificate_authentication')
+        if version == Version.TLS12 and side == Side.CLIENT:
+            #The server uses an ECDSA cert, so make sure we have a compatible key exchange
```
mpg, line 94: Style: space after `#`

scripts/generate_tls_handshake_tests.py
```diff
@@ -0,0 +1,175 @@
...
+    out.write("""\
+# End of automatically generated file.
+""")
+
+def main() -> None:
+    """Command line entry point."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-o', '--output',
+                        default='tests/opt-testcases/handshake-generated.sh',
+                        help='Output file')
```
mpg, line 169: Minor: when I invoke the script with `--help` I'd like to show the default value of the output file, but that's not happening at the moment.

mpg (approved):
Looks good to me on code review (the only feedback I have is minor enough that it can be considered optional). I've also reviewed the diff of the output and the current content of `ssl-opt.sh` and I'm satisfied that all the differences are OK.

scripts/generate_tls_handshake_tests.py
```diff
@@ -0,0 +1,175 @@
...
+            tc.requirements.append('requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3')
+            tc.description += '  TLS 1.3 ClientHello -> 1.2 Handshake'
+
+    # To guarantee that the handhake messages are large enough and need to be
+    # split into fragments, the tests require certificate authentication.
+    # The party in control of the fragmentation operations is OpenSSL and
+    # will always use server5.crt (548 Bytes).
+    if length is not None and \
+       length >= TLS_HANDSHAKE_FRAGMENT_MIN_LENGTH:
+        tc.requirements.append('requires_certificate_authentication')
```
mpg, line 92: I think if we move this out of the `if` we won't need the extra changes in the TLS PR that [I'm complaining about](https://github.com/Mbed-TLS/mbedtls/pull/10021#discussion_r1979233633).

mpg (changes requested):
Sorry, changed my mind after reviewing #10021.

Another way to look at it: the script is supposed to generate something equivalent to what we had manually written. If it does, we shouldn't need the two controversial commits in #10021 because we didn't need them with the manually written tests.

scripts/generate_tls_handshake_tests.py
mpg: For the record: I think I was a victim of XYZ syndrome with the above comment: I want X (avoid changing OpenSSL default options in a way that impacts non-defragmentation tests), and instead I'm asking for Y (restore this arguably too-strong dependency) while perhaps there's a better way Z (add an explicit `-allow_no_dhe_kex` in the defragmentation tests, which would not only address X but also give us some defragmentation testing in PSK-only configs as a welcome bonus).

minosgalanakis (commented):
Overall it does what expected, and given the time contrains when this PR was introduced I appeciate that it should't need a complicated design review.

All my comments are indicative and only consider them if reworking it and they make sense.

scripts/generate_tls_handshake_tests.py
gilles-peskine-arm: I've changed to injecting `-allow_no_dhe_kex` where it's needed (or at least a superset of that which remains a subset of the handshake defragmentation test cases).

scripts/generate_tls_handshake_tests.py
```diff
@@ -167,9 +167,10 @@ def write_handshake_tests(out: typing_util.Writable) -> None:
 def main() -> None:
     """Command line entry point."""
     parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument('-o', '--output',
                         default='tests/opt-testcases/handshake-generated.sh',
-                        help='Output file')
+                        help='Output file (default: tests/opt-testcases/handshake-generated.sh)')
```
minosgalanakis, line 173: Love the simplicity of this one ;)

scripts/generate_tls_handshake_tests.py
gilles-peskine-arm: Filed https://github.com/Mbed-TLS/mbedtls-framework/issues/144 for maintainability improvements.

gilles-peskine-arm:
The CI is now failing because https://github.com/Mbed-TLS/mbedtls-framework/pull/140 and its 3.6 consumer have just been merged :frowning_face: 

Given the nature of the changes here, the CI should pass on the merge of this pull request with `main`. However it'll break when we update the framework in the `features/tls-defragmentation/*` branches.

@mpg @minosgalanakis What do you think — another update of the consuming branches? Or should we create a feature branch in the framework as well?

minosgalanakis:
> The CI is now failing because #140 and its 3.6 consumer have just been merged ☹️
> 
> Given the nature of the changes here, the CI should pass on the merge of this pull request with `main`. However it'll break when we update the framework in the `features/tls-defragmentation/*` branches.
> 
> @mpg @minosgalanakis What do you think — another update of the consuming branches? Or should we create a feature branch in the framework as well?

Looking at the log

~~~~~~
[2025-03-04T18:37:57.204Z] ../library/libmbedcrypto.a(md.o): In function `mbedtls_md':
[2025-03-04T18:37:57.204Z] md.c:(.text+0x666): undefined reference to `psa_can_do_hash'
[2025-03-04T18:37:57.204Z] md.c:(.text+0x69c): undefined reference to `psa_hash_compute'
[2025-03-04T18:37:57.204Z] clang: error: linker command failed with exit code 1 (use -v to see invocation)
[2025-03-04T18:37:57.204Z] Makefile:159: recipe for target 'aes/crypt_and_hash' failed
[2025-03-04T18:37:57.204Z] make[1]: *** [aes/crypt_and_hash] Error 1
~~~~~~

It appears like the CI failed because the tf-psa-framework pointer is at a different state to what it should be?

#142 as it stands it adds new files which nobody else is using so as you say could be safely merged into framework:main. I cannot confidently guess as to why that is not working.

That bein said, given the time contrains I would prefer we move the files to the main mbedtls, review and merge them into the existing feature branch, and after the feature branch has been merged we can move it onto the framework.

But that is my opinion only ;)

gilles-peskine-arm:
I've gone and created the branch.

The CI on the framework will still pick `mbedtls-3.6` and `development`. We'll just have to ignore it and look at the CI on consuming PR.

minosgalanakis (approved):
LGTM

gilles-peskine-arm:
CI status: the CI has passed on https://github.com/Mbed-TLS/mbedtls/pull/10029 and https://github.com/Mbed-TLS/mbedtls/pull/10021, so this is ok to merge in the `tls-defragmentation` feature branch.

PR #153, opened 2025-04-02
Move everest headers

bjwtaylor:
## Description

Move everest headers

## PR checklist

- [ ] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/235
- [ ] **development PR** provided #HERE
- [ ] **3.6 PR** not required because: No backports

### DIFF
--- a/scripts/test_psa_constant_names.py
+++ b/scripts/test_psa_constant_names.py
@@ -173,7 +173,7 @@ def main():
                             action='append', default=['tf-psa-crypto/include',
                                                       'tf-psa-crypto/drivers/builtin/include',
                                                       'tf-psa-crypto/drivers/everest/include',
-                                                      'tf-psa-crypto/drivers/everest/include/'
+                                                      'tf-psa-crypto/drivers/everest/include/' +
                                                       'tf-psa-crypto/private',
                                                       'include'],
                             help='Directory for header files')


scripts/test_psa_constant_names.py
```diff
@@ -172,7 +172,8 @@ def main():
         parser.add_argument('--include', '-I',
                             action='append', default=['tf-psa-crypto/include',
                                                       'tf-psa-crypto/drivers/builtin/include',
-                                                      'tf-psa-crypto/drivers/everest/include',
+                                                      'tf-psa-crypto/drivers/everest/include/',
```
felixc-arm, line 175: Has this change accidentally been split on to two lines? I assume it's meant to be `tf-psa-crypto/drivers/everest/include/tf-psa-crypto/private` rather than two separate elements in the list.

scripts/check_names.py
```diff
@@ -702,8 +702,8 @@ def comprehensive_parse(self):
             "include/psa/*.h",
             "include/tf-psa-crypto/*.h",
             "drivers/builtin/include/mbedtls/*.h",
-            "drivers/everest/include/everest/everest.h",
```
felixc-arm, line 705: I also noticed in this file that there are `3rdparty/everest/include/everest/<everest.h|x25519.h>` paths, I'm not sure if they'll need changing too or not...

felixc-arm (changes requested):

scripts/test_psa_constant_names.py
bjwtaylor: It just looks like a typo, I've removed it.

scripts/check_names.py
bjwtaylor: I've removed these lets see if the CI passes.

felixc-arm: I was thinking more along the lines of changing them to the new path, e.g. `3rdparty/everest/include/tf-psa-crypto/private/everest/<everest.h|x25519.h>` rather than removing them - removing them does seem to be causing the build errors it seems.

Anyway/Also after having another look it seems this part is only executed for 3.6 and so naturally we don't want to change 3.6-related stuff and so this `3rdparty/` code should be left as it was... apologies for bringing it up :melting_face:

bjwtaylor: OK, I'll revert the changes.

scripts/test_psa_constant_names.py
felixc-arm: I think you just missed the `/` after `include` (or before the second `tf-psa-crypto`) in the updated patch - although I don't think the CI will pass here anyway as it's using the development `mbedtls` & `tf-psa-crypto` so naturally can't find anything in `tf-psa-crypto/drivers/everest/include/tf-psa-crypto/private` as the directory doesn't exist.

I this issue (the directory doesn't exist in development) is causing both the CI issues (i.e. the check_names issue and the compilation issue where it tries to include `drivers/everest/include/tf-psa-crypto/private/` from the change in `psa_stoage.py`) - so _presumably_ we would want this patch changed to accommodate both scenarios (everest headers are in `drivers/everest/include` _or_ `drivers/everest/include/tf-psa-crypto/private`) and then can remove the old functionality in a later patch...

gilles-peskine-arm (changes requested):
This breaks (or apparently just removes some checks from) 3.6.

scripts/test_psa_constant_names.py
bjwtaylor: I think this is a duplicate @gilles-peskine-arm comment. Basically we will need both the 3.6 paths and the new path present or the CI will fail with the older version. I'll resolve the one below and hopefully that will resolve both.

scripts/test_psa_constant_names.py
```diff
@@ -173,6 +173,7 @@ def main():
                             action='append', default=['tf-psa-crypto/include',
                                                       'tf-psa-crypto/drivers/builtin/include',
                                                       'tf-psa-crypto/drivers/everest/include',
+                                                      'tf-psa-crypto/private',
```
felixc-arm, line 176: LGTM overall, however `tf-psa-crypto/private` isn't a path that exists so I don't think it needs to be here.

Presumably we also need the new path here as well?

bjwtaylor: Now added

felixc-arm: Shouldn't it be `tf-psa-crypto/drivers/everest/include/tf-psa-crypto/private` rather than `drivers/everest/include/tf-psa-crypto/private/everest`?

felixc-arm: I think you forgot the `tf-psa-crypto/` at the start here...

The line's also now too long, so I guess
```
'tf-psa-crypto/drivers/everest/include/' +
'tf-psa-crypto/private',
```
Would work.

felixc-arm (approved):
LGTM, cheers!

scripts/check_names.py
```diff
@@ -704,7 +704,9 @@ def comprehensive_parse(self):
             "include/mbedtls/*.h",
             "drivers/builtin/include/mbedtls/*.h",
             "drivers/everest/include/everest/everest.h",
-            "drivers/everest/include/everest/x25519.h"
+            "drivers/everest/include/everest/x25519.h",
+            "drivers/everest/include/tf-psa-crypto/private/everest/everest.h",
+            "drivers/everest/include/tf-psa-crypto/private/everest/x25519.h"
```
gilles-peskine-arm, line 709: Non-blocker: we should put a comma after each list item, even the last one. This would avoid having to change the last element when adding another item after it. We usually do that, this file was one of the minority that doesn't.

gilles-peskine-arm (approved):
LGTM

PR #123, opened 2025-01-13
Adapt all-helpers.sh ARMC6 build for TF-PSA-Crypto

Harry-Ramsey:
This commit adapts helper_armc6_build_test in all-helpers.sh to build for both MbedTLS using make and TF-PSA-Crypto using CMake.

## PR checklist

Please add the numbers (or links) of the associated pull requests for consuming branches. You can omit branches where this pull request is not needed.

- [ ] **crypto PR** Mbed-TLS/TF-PSA-Crypto: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/155
- [ ] **development PR** Mbed-TLS/mbedtls#
- [ ] **3.6 PR** Mbed-TLS/mbedtls#

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

Help make review efficient:
* Multiple simple commits
  - please structure your PR into a series of small commits, each of which does one thing
* Avoid force-push
  - please do not force-push to update your PR - just add new commit(s)
* See our [Guidelines for Contributors](https://mbed-tls.readthedocs.io/en/latest/reviews/review-for-contributors/) for more details about the review process.

PR #154, opened 2025-04-04
[framwork] Add components-compiler.sh

valeriosetti:
## Description

Helps resolving https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/60

It supersedes https://github.com/Mbed-TLS/mbedtls-framework/pull/123

## PR checklist

- [ ] **TF-PSA-Crypto PR** 
- [ ] **development PR** not required
- [ ] **3.6 PR** not required


gabor-mezei-arm (approved):
LGTM

scripts/all-helpers.sh
```diff
@@ -258,17 +258,30 @@ helper_armc6_build_test()
...
+            "$ARMC6_FROMELF" -z ${PSA_CORE_PATH}/*.o
+        fi
+        if [ -n "${BUILTIN_SRC_PATH}" ]; then
+            "$ARMC6_FROMELF" -z ${BUILTIN_SRC_PATH}/*.o
+        fi
+    else
+        cd $TF_PSA_CRYPTO_ROOT_DIR
+        rm -rf $OUT_OF_SOURCE_DIR
+        mkdir $OUT_OF_SOURCE_DIR
+        cd $OUT_OF_SOURCE_DIR
```
ronald-cron-arm, line 278:
```suggestion
```
I would not do that here and let the callers do it if necessary.

scripts/all-helpers.sh
```diff
@@ -258,17 +258,30 @@ helper_armc6_build_test()
...
+        cmake -DCMAKE_SYSTEM_NAME="Generic" -DCMAKE_SYSTEM_PROCESSOR="cortex-m0" \
+              -DCMAKE_C_COMPILER="$ARMC6_CC" -DCMAKE_C_LINKER="$ARMC6_LINK" \
+              -DCMAKE_AR="$ARMC6_AR" -DCMAKE_C_FLAGS="$FLAGS" \
+              -DCMAKE_C_COMPILER_WORKS=TRUE -DENABLE_TESTING=OFF \
+              -DENABLE_PROGRAMS=OFF "$TF_PSA_CRYPTO_ROOT_DIR"
+        make
```
ronald-cron-arm, line 284: What about the size part, calls to "$ARMC6_FROMELF" -z */*.o?

ronald-cron-arm (changes requested):
I have a few comments.

gabor-mezei-arm (dismissed):
LGTM

tests/programs/test_zeroize.gdb
```diff
@@ -33,17 +33,19 @@
...
 set confirm off
 
+# TF-PSA-Crypto uses a different name for the executable and also a build out
+# of tree.
 if $_isvoid($is_tf_psa_crypto)
     file ./programs/test/zeroize
+    set args ./framework/tests/programs/zeroize.c
 else
     file ./programs/test/tf_psa_crypto_zeroize
+    set args ../framework/tests/programs/zeroize.c
```
ronald-cron-arm, line 43: If I understand correctly the ".." is because of the out-of-tree build which is not really something we should care about here. Thus I would suggest to rather build in tree in `test_tf_psa_crypto_zeroize`.

ronald-cron-arm (changes requested):

ronald-cron-arm (approved):
LGTM, thanks.

Issue #157, opened 2025-04-10
Make demo_common.sh usable on its own


gilles-peskine-arm:
Since https://github.com/Mbed-TLS/mbedtls-framework/pull/141, `framework/scripts/demo_common.sh` uses functions from `framework/scripts/project_detection.sh`. But it doesn't source `project_detection.sh`: that's up to its caller. This defeats part of the purpose of `demo_common.sh`, which is to minimize boilerplate in demo scripts, which are intended to be easy to follow for users who want to understand what the script is doing.

This has broken existing demo scripts — see https://github.com/Mbed-TLS/mbedtls/pull/10129.

https://github.com/Mbed-TLS/mbedtls/pull/10129 is a quick fix but it is not really satisfactory. If `demo_common.sh` needs `project_detection.sh`, it should source it itself. However there is a bootstrap problem: `demo_common.sh` only knows the location of the original script (`$0`), not its own location (this is not available in plain sh). So `demo_common.sh` needs to locate the framework submodule somehow. Since `demo_common.sh` already has a loop to locate the root directory (strangely, if it assumes that `project_detection.sh` is available), we could use this loop to find the framework.

What the loop does with respect to mbedtls vs tf-psa-crypto is also a bit strange, see https://github.com/Mbed-TLS/mbedtls/pull/10129#discussion_r2037307799. This may or may not be in scope here.

Definition of done: `demo_common.sh` can be included on its own, and does all the dirty work of locating directories and helper tools.

valeriosetti:
Re-opening because https://github.com/Mbed-TLS/mbedtls/pull/10132 is still in the merge queue. When that one will be also merged this issue will be completely closed.

PR #159, opened 2025-04-11
[framework] Make demo_common.sh usable on its own

valeriosetti:
## Description

It helps resolving https://github.com/Mbed-TLS/mbedtls-framework/issues/157

## PR checklist

- [ ] **TF-PSA-Crypto PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/255
- [ ] **development PR** not requried
- [ ] **3.6 PR** https://github.com/Mbed-TLS/mbedtls/pull/10132

### DIFF
--- a/scripts/demo_common.sh
+++ b/scripts/demo_common.sh
@@ -16,14 +16,13 @@
 
 set -e -u
 
-DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG=${DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG:-1}
-
-need_query_compile_time_config () {
-  if [ $DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG -eq 1 ]; then
-    return 0;
-  else
+# Check if the provided path ($1) can be a valid root for Mbed TLS or TF-PSA-Crypto.
+# This is based on the fact that "scripts/project_name.txt" exists.
+is_valid_root () {
+  if ! [ -f "$1/scripts/project_name.txt" ]; then
     return 1;
   fi
+  return 0;
 }
 
 ## At the end of the while loop below $root_dir will point to the root directory
@@ -35,21 +34,15 @@ root_dir="${0%/*}"
 ##
 ## The code works no matter where the demo script is relative to the current
 ## directory, even if it is called with a relative path.
-n=5
+n=4
 while true; do
   # If we went up too many folders, then give up and return a failure.
   if [ $n -eq 0 ]; then
     echo >&2 "This doesn't seem to be an Mbed TLS source tree."
     exit 125
   fi
-  # If we reached the Mbed TLS root folder then we're done.
-  if is_mbedtls_root "$root_dir"; then
-    break;
-  fi
-  # If we reached the TF-PSA-Crypto root folder and the script that sourced
-  # this file does not need query_compile_time_config (which is only available
-  # in Mbed TLS repo) then we're done.
-  if is_tf_psa_crypto_root "$root_dir" && ! need_query_compile_time_config; then
+
+  if is_valid_root "$root_dir"; then
     break;
   fi
 
@@ -62,6 +55,9 @@ while true; do
     *) root_dir=".";;
   esac
 done
+
+# Now that we have a root path we can source the "project_detection.sh" script.
+. "$root_dir/framework/scripts/project_detection.sh"
 
 ## msg LINE...
 ## msg <TEXT_ORIGIN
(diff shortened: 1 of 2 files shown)


valeriosetti:
> I'm still not sure what `project_detection.sh` is doing here. Ok, we need `query_compile_time_config`, but `demo_common.sh` doesn't rely on `project_detection.sh` to determine whether it's available.

Yes, `demo_common.sh` does not need `project_detection.sh`, but `dlopen_demo.sh` [does](https://github.com/Mbed-TLS/mbedtls-framework/blob/main/tests/programs/dlopen_demo.sh#L20). As far as I understood `demo_common.sh` should be the only thing other scripts need to source in order to have all the things they need, so that's why `project_detection.sh` ended up here.

scripts/demo_common.sh
```diff
@@ -63,6 +77,9 @@ while true; do
   esac
 done
 
+# Now that we have a root path we can source the "project_detection.sh" script.
```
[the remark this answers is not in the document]
valeriosetti: `demo_common.sh` does not need `project_detection.sh` directly, but `dlopen_demo.sh` [does](https://github.com/Mbed-TLS/mbedtls-framework/blob/main/tests/programs/dlopen_demo.sh#L20). As far as I understood `demo_common.sh` should be the only thing other scripts need to source in order to have all the things they need, so that's why `project_detection.sh` ended up here.

scripts/demo_common.sh
```diff
@@ -42,15 +54,17 @@ while true; do
...
-    break;
+
+  if is_valid_root "$root_dir"; then
+    if ! need_query_compile_time_config; then
+      break;
+    fi
+    # query_compile_time_config is in the Mbed TLS repo, not in the TF-PSA-Crypto
+    # one, so if we reached the TF-PSA-Cryto root path then we might need to
+    # step one folder ahead.
+    if need_query_compile_time_config && has_query_compile_time_config "$root_dir"; then
```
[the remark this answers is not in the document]
valeriosetti: This check is for when `TF-PSA-Crypto` is not a subproject of `Mbed TLS`. In that case returning the parent folder wouldn't automatically solve the problem. To be more generic and not making any assumption I think it's better to just continue moving one folder ahead in that case and see if that one is OK (i.e. the current implementation). Wdyt?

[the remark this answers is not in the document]
gilles-peskine-arm: I don't understand the difference between “returning the parent folder” and “continue moving one folder ahead”.

scripts/demo_common.sh
```diff
@@ -63,6 +80,9 @@ while true; do
   esac
 done
 
+# Now that we have a root path we can source the "project_detection.sh" script.
+. "$root_dir/framework/scripts/project_detection.sh"
```
gilles-peskine-arm, line 84: I'm not familiar with the history here. Have you considered having a `demo_common.sh` in each branch that sources a `demo_framework.sh` in the framework?

scripts/demo_common.sh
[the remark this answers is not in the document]
valeriosetti: If we just return the parent folder, but the parent folder is not the Mbed TLS repo, then `query_compile_time_config` won't be available. So yes, I can return the parent folder, but then after the while loop I need to the check for `need_query_compile_time_config && has_query_compile_time_config "$root_dir"`.
If instead I let the while loop continue its travel to the parent folder it will fail without requiring any extra check. Am I wrong?

scripts/demo_common.sh
valeriosetti: Do you mean having a `demo_common.sh` in `tf-psa-crypto/programs/psa` for `key_ladder_demo` and `psa_hash_demo` and one in `frameworks/tests/programs` for `dlopen_demo`?
That's an option. The only drawback I see is the partial duplication of the initial `while` loop to search for the root folder. Not full duplication because there would be differences, but is it worth it?
If you prefer that way just let me know and I will implement this path ;)

ronald-cron-arm: I don't really understand what is the proposition here and what problem we are trying to solve.

gilles-peskine-arm: We have:

1. Demo scripts in consuming branches.
2. Common code to all demo scripts.
3. Code that wants to know which branch it's consumed by.

Currently the common code is in `framework/scripts/demo_common.sh`, and it has somewhat complicated code to both detect the project root and identify the branch, with some special casing because some scripts want to find mbedtls-parenting-crypto when run from a crypto script.

My alternative proposal is to have an intermediate script in each branch that does the branch-specific setup, locates the framework and calls the common code in the framework. This way, no branch detection is necessary.

ronald-cron-arm: Okay but with the proposal in https://github.com/ronald-cron-arm/mbedtls-framework/tree/pr159-test based on [one of your proposal](https://github.com/Mbed-TLS/mbedtls-framework/pull/159#discussion_r2040101717) (or at least my interpretation of it) the project root detection is not branch specific anymore. We have only one temporary trick for the path to `query_compile_time_config` that we can remove when support for `query_compile_time_config` is added in TF-PSA-Crypto. We have https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/127 tracking that work.

valeriosetti: @ronald-cron-arm I just updated my commit including changes from your branch.

scripts/demo_common.sh
```diff
@@ -105,18 +101,33 @@ run_bad () {
...
+else
+  query_compile_time_config_dir=""
+fi
+
 ## config_has SYMBOL...
 ## Succeeds if the library configuration has all SYMBOLs set.
 ##
 ## Note: "query_compile_time_config" is only available when in Mbed TLS project,
 ## so "config_has" is not available in TF-PSA-Crypto one. If this function is
 ## called from the latter we fail immediately.
```
gilles-peskine-arm, line 124: If `query_compile_time_config` is needed but absent (for example, if the executable was built out of tree), the demo script will silently be skipped. Is this really what we want?

gilles-peskine-arm (commented):
It looks wrong to me that if `query_compile_time_config` is expected to be present but is not found, then the demo script will silently be skipped. Is this intended? Is it documented somewhere? Do we have a plan to fix it?

LGTM apart from this.

ronald-cron-arm (dismissed):
LGTM, thanks.

gilles-peskine-arm (changes requested):
Now the script emits a message if `query_compile_time_config` is missing, but it still doesn't error out.

scripts/demo_common.sh
```diff
@@ -105,18 +101,33 @@ run_bad () {
...
+
 ## config_has SYMBOL...
 ## Succeeds if the library configuration has all SYMBOLs set.
 ##
-## Note: "query_compile_time_config" is only available when in Mbed TLS project,
-## so "config_has" is not available in TF-PSA-Crypto one. If this function is
-## called from the latter we fail immediately.
+## Note: depending on the above check query_compile_time_config_dir might be
+##       intentionally set to "". In this case the call to
+##       query_compile_time_config below will make the script to fail.
```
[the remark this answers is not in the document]
valeriosetti: > By the way, I notice that demo_common.sh itself calls config_has, so there was no need for DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG: it's effectively always true.

That was a flag used by other scripts sourcing `demo_comomn.sh`: `psa_hash_demo` and `jey_ladder_demo` were setting `DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG=1`, while `dlopen_demo` was setting `DEMO_COMMON_NEED_QUERY_COMPILE_TIME_CONFIG=0`. Probably the name was not the best, but that symbol made sf ense in the original state of this script (prior to this PR), but now it became useless.

> In if has_config …, the effect of set -e is cancelled for all the code involved in the if condition, so the failing call to query_compile_time_config does not cause the for loop to exit, and then the if condition can only distinguish between success and failure.

Oh, that's tricky, but good to know. Thanks for the tip!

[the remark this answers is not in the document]
valeriosetti: > It would be a lot easier to just check whether query_compile_time_config exists in demo_common.sh, since it's always used.

I don't think it's OK to make `demo_common` to always fail if `query_compile_time_config` is not available. When tf-psa-crypto is used as standalone project (i.e. not inside the Mbed TLS one) there won't be any `query_compile_time_config` but we might want to run `dlopen_demo` anyway, It's perhaps better to improve `depends_on` IMO

ronald-cron-arm (approved):
LGTM, the OpenCI failure is due to the failure of a ssl-opt.sh test, an instance of https://github.com/Mbed-TLS/mbedtls/issues/8471 it seems.

gilles-peskine-arm (approved):
Thanks for updating, looks good to me now.

ronald-cron-arm (approved):
LGTM, thanks.

PR #160, opened 2025-04-14
remove compat-2.x.h

bjwtaylor:
## Description

Remove include/mbedtls/compat-2.x.h and its reference in check_names.py.

## PR checklist

- [ ] **TF-PSA-Crypto PR** not required because: No changes
- [ ] **development PR** provided: #here
- [ ] **3.6 PR** not required because: No backports

### DIFF
--- a/scripts/check_names.py
+++ b/scripts/check_names.py
@@ -223,7 +223,7 @@ class CodeParser():
 
         # Globally excluded filenames.
         # Note that "*" can match directory separators in exclude lists.
-        self.excluded_files = ["*/bn_mul", "*/compat-2.x.h"]
+        self.excluded_files = ["*/bn_mul"]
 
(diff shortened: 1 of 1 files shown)


felixc-arm (commented):
This'll break 3.6 CI right? So presumably the main `mbedtls` change isn't suitable for a backport as it'll break the API (by removing backwards compatibility for 3.6 vs 2.x?), so this change shouldn't be made (or should be guarded so that the `compat-2.x.h` exclusion is present for 3.6 and removed for 4.x).

Unless of course I'm mistaken and the main `mbedtls` PR should be backported...

bjwtaylor:
@felixc-arm, this is a good point. Closing this PR as it will not be required as we can leave the path included and it will just not be found in later versions.

PR #164, opened 2025-04-25
Mechanism to error out on removed configuration options

gilles-peskine-arm:
Set up a mechanism for erroring out if users try to use removed options in their configuration file. Resolves https://github.com/Mbed-TLS/mbedtls/issues/10147.

Status: work in progress. This will be rebased. For the time being, please review only the logic of what is allowed and rejected, not the implementation.

## PR checklist

- [x] **TF-PSA-Crypto PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/269
- [ ] **development PR** TODO
- [x] **3.6 PR** not required because: new code for use in 1.0/4.0 only


scripts/mbedtls_framework/config_history.py
```diff
@@ -0,0 +1,57 @@
...
+            assert m is not None
+            (product, version) = m.groups()
+            options = self.load_options(filename)
+            exclude = exclude_dict.get(product, {}).get(version, frozenset())
+            data.setdefault(product, {})[version] = options - exclude
+        return data
+
+    def __init__(self) -> None:
+        """Load all files containing historical option lists."""
+        self._history_dir = os.path.join(build_tree.framework_root(), 'history')
```
gilles-peskine-arm, line 47: One thing I think I didn't document properly, so I'll at least leave a note here: why are the historical option lists in the framework, rather than in each consuming project? After all, you'd normally expect each project to know about itself.

First of all, I didn't want running tests to ever need to access another branch. Checking for removed options requires knowledge about two older commits: the end of the previous major version, and the beginning of the current major version. This historical knowledge is stored in the branch under test (including its submodules) so that the branch can be built and tested on its own.

Mbed TLS needs to know about internal options of crypto, so that it doesn't complain if crypto internally sets an option that was public in 3.6 but removed in 1.0/4.0. This could be communicated upwards. But crypto also needs to know about public options of the new version of Mbed TLS, so that it doesn't complain if `crypto_config.h` tries to set an X.509 or TLS option: we want to allow users to just have a single config file that's used as both the crypto config and the mbedtls config. So projects need information to propagate upwards as well as downwards. Since the crypto build doesn't have access to an mbedtls tree, the information about current Mbed TLS needs to be in crypto or in the framework.

Having information about Mbed TLS in crypto doesn't make a lot of sense, and is just as bad in terms of putting things in the wrong project as having that information in the framework. So I put everything in the framework. This way at least it's in one place.

This does mean that we'll sometimes want to update the framework because of a change in a superproject. But this will only be needed when we remove an option, and that normally only ever happens while we're preparing a major version. For 1.0/4.0, because the checks came in so late before the release, we'll probably do just one update of the 1.0/4.0 data after the release. For 2.0/5.0, if this mechanism still exists in its current form, we'll probably do a few more updates, or maybe we'll turn off the 3.6-to-1.0/4.0 checks first and then only add 1.0-to-2.0 checks close to the 2.0 release.

PR #167, opened 2025-05-07
[framework] Remove MBEDTLS_USE_PSA_CRYPTO from PK module

valeriosetti:
## Description

This helps resolving https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/283

## PR checklist

- [ ] **TF-PSA-Crypto PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/283
- [ ] **development PR** no
- [ ] **3.6 PR** no


bjwtaylor (dismissed):

tests/src/psa_exercise_key.c
```diff
@@ -1430,7 +1430,10 @@ int mbedtls_test_key_consistency_psa_pk(mbedtls_svc_key_id_t psa_key,
             break;
 #endif /* PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY && !MBEDTLS_PK_USE_PSA_EC_DATA */
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    /* In TF-PSA-Crypto USE_PSA is always enabled. However we cannot
+     * remove this guard because this test code can be used also with
+     * mbedtls-3.6 branch. */
+#if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(PSA_CRYPTO_API_VERSION_MAJOR)
```
mpg, line 1436: I don't think this is incorrect, but I don't understand why it would be needed now. Unless I'm missing something, `MBEDTLS_USE_PSA_CRYPTO` is still being defined in tf-psa-crypto for now (in `https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/278`) and still needs to be until we finish getting rid of it on the TLS side.

At some point in the future we'll stop defining it, and then I agree we'll need a change like this one, but in the meantime I don't understand why we'd need this. Am I missing something?

mpg (commented):
I don't understand why this is needed now :)

tests/src/psa_exercise_key.c
valeriosetti: > At some point in the future we'll stop defining it, and then I agree we'll need a change like this one, but in the meantime I don't understand why we'd need this. Am I missing something?

No you are not. I was just thinking about this future case and I thought it could be nice to "prepare" framework for it. Since it's not needed I will remove it.

PR #175, opened 2025-06-16
  Move private headers to a private directory

amtkarm1:
## Description

Fixes https://github.com/Mbed-TLS/mbedtls/issues/10087.

## PR checklist

- [X] **TF-PSA-Crypto PR** provided Mbed-TLS/TF-PSA-Crypto#318
- [X] **development PR** provided Mbed-TLS/mbedtls#10224
- [X] **3.6 PR** not required because: no backport required

### DIFF
--- a/tests/include/test/bignum_helpers.h
+++ b/tests/include/test/bignum_helpers.h
@@ -17,7 +17,11 @@
 
 #if defined(MBEDTLS_BIGNUM_C)
 
+#if MBEDTLS_VERSION_MAJOR >= 4
 #include <mbedtls/private/bignum.h>
+#else
+#include <mbedtls/bignum.h>
+#endif
 #include <bignum_mod.h>
 
 /** Allocate and populate a core MPI from a test case argument.
--- a/tests/include/test/bignum_helpers.h
+++ b/tests/include/test/bignum_helpers.h
@@ -17,7 +17,7 @@
 
 #if defined(MBEDTLS_BIGNUM_C)
 
-#if MBEDTLS_VERSION_MAJOR >= 4
+#if !defined(MBEDTLS_VERSION_MAJOR) || MBEDTLS_VERSION_MAJOR >= 4
 #include <mbedtls/private/bignum.h>
 #else
 #include <mbedtls/bignum.h>
--- a/tests/include/test/bignum_helpers.h
+++ b/tests/include/test/bignum_helpers.h
@@ -17,7 +17,7 @@
 
 #if defined(MBEDTLS_BIGNUM_C)
 
-#include <mbedtls/bignum.h>
+#include <mbedtls/private/bignum.h>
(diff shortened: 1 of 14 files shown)


tests/include/test/bignum_helpers.h
```diff
@@ -17,7 +17,7 @@
 
 #if defined(MBEDTLS_BIGNUM_C)
 
-#include <mbedtls/bignum.h>
+#include <mbedtls/private/bignum.h>
```
mpg, line 20: This is good for development, but the framework needs to keep working with 3.6 as well. I suggest
```c
#if MBEDTLS_VERSION_MAJOR >= 4
#include <mbedtls/private/bignum.h>
#else
#include <mbedtls/bignum.h>
#endif
```
(Here and in all similar places obviously.) (Edit: `MBEDTLS_VERSION_MAJOR` is defined by including `build_info.h` which should already be included everywhere.)

I also suggest you create a "shadow" 3.6 PR where the only thing you do is change the framework pointer to the head of this PR.

Your goal is now to get a green CI both on the mbedtls development PR _and_ the shadow 3.6 PR :)

mpg (changes requested):

tests/include/test/bignum_helpers.h
gilles-peskine-arm: Not `PSA_CRYPTO_API_VERSION_xxx`: that's the API version which is irrelevant. We'd need `TF_PSA_CRYPTO_VERSION_xxx` (or whatever the name is supposed to be) which [doesn't exist yet](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/323).

valeriosetti: OK, thanks for the details! So should we file a follow-up issue for whenever the tf-psa-crypto's symbols corresponding to `MBEDTLS_VERSION_MAJOR` will be defined? Because I suspect CI tests for standalone tf-psa-crypto here are passing just because the old headers' path is picked. Indeed there are tests on https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/318 which are failing for this reason (I didn't check them all, but at least `test_tf_psa_crypto_out_of_source` does):
```
[2025-06-24T10:46:42.522Z] [  1%] Building C object CMakeFiles/tf_psa_crypto_test.dir/framework/tests/src/bignum_helpers.c.o
[2025-06-24T10:46:42.522Z] In file included from /var/lib/build/framework/tests/src/bignum_helpers.c:14:
[2025-06-24T10:46:42.522Z] /var/lib/build/framework/tests/include/test/bignum_helpers.h:23:10: fatal error: 'mbedtls/bignum.h' file not found
[2025-06-24T10:46:42.522Z] #include <mbedtls/bignum.h>
```

davidhorstmann-arm (approved):
LGTM, thanks!

PR #183, opened 2025-07-10
Remove legacy crypto options preparation-1

ronald-cron-arm:
## Description
Some changes to prepare for the removal of legacy crypto options. The changes should be enough to remove hash, cipher, cmac and aead configuration options in TF-PSA-Crypto with a green CI.

The first commit is a follow-up of https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/332. I've put it here to reduce the number of PRs and CI runs.

## PR checklist
- [x] **TF-PSA-Crypto PR** not required 
- [x] **development PR** not required 
- [x] **3.6 PR** not required

### DIFF
--- a/scripts/generate_config_tests.py
+++ b/scripts/generate_config_tests.py
@@ -68,6 +68,13 @@ SIMPLE_DEPENDENCIES = {
 if build_tree.is_mbedtls_3_6():
     SIMPLE_DEPENDENCIES['MBEDTLS_NO_PLATFORM_ENTROPY'] = 'MBEDTLS_ENTROPY_C'
 
+BUILTIN_MODULE_ENABLEMENT_MACROS = [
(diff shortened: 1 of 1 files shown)


minosgalanakis (approved):
LGTM with regards to the logic of the changes.

Looking at the CI failure, it failed at a single spot at the `./tests/scripts/all.sh --seed 4 --keep-going test_no_renegotiation` when `server anti-replay off` with `pattern 'resend' MUST be present in the Client output`

I cannot see how the changes in the PR could have caused that and it could be intermittent but should be investigated. Also are we supporting TLS 1.2 renegotiation in 1.0.0?

ronald-cron-arm:
> Looking at the CI failure, it failed at a single spot at the `./tests/scripts/all.sh --seed 4 --keep-going test_no_renegotiation` when `server anti-replay off` with `pattern 'resend' MUST be present in the Client output`
> 
> I cannot see how the changes in the PR could have caused that and it could be intermittent but should be investigated. 

The "New CI" has run successfully thus it's likely to be a CI glitch.

PR #193, opened 2025-07-30
Exclude "WIN32_LEAN_AND_MEAN" from name check

irwir:
## Description

Macro `WIN32_LEAN_AND_MEAN` excludes a few APIs that are rarely used. If needed, the necessary headers would be added explicitly.
The macro goes into a header file and thus should be in the exceptions of the name check script (to be used in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/262 or a follow-up).

## PR checklist

- [x] **TF-PSA-Crypto PR** not required
- [x] **development PR** not required
- [x] **3.6 PR** not required because no backport was planned.


gilles-peskine-arm (approved):
LGTM

PR #197, opened 2025-08-05
[framework] Remove 224-bit curves (1/3)

valeriosetti:
## Description

It helps resolving https://github.com/Mbed-TLS/mbedtls/issues/10283
It's PR 1 out of 2

## PR checklist

- [ ] **TF-PSA-Crypto PR** provided: TODO
- [ ] **development PR** provided: TODO 
- [ ] **3.6 PR** not required because: not backported

### DIFF
--- a/scripts/mbedtls_framework/ecp.py
+++ b/scripts/mbedtls_framework/ecp.py
@@ -7,6 +7,7 @@ from typing import List
 
 from . import test_data_generation
 from . import bignum_common
+from . import build_tree
 
 
 class EcpTarget(test_data_generation.BaseTarget):
@@ -164,7 +165,9 @@ class EcpP224R1Raw(bignum_common.ModOperationCommon,
 
     @property
     def is_valid(self) -> bool:
-        return True
+        # secp224r1 support has been removed from development, but it's stil
+        # available in 3.6 branch.
+        return build_tree.is_mbedtls_3_6()
 
     def arguments(self)-> List[str]:
         args = super().arguments()
@@ -624,7 +627,9 @@ class EcpP224K1Raw(bignum_common.ModOperationCommon,
 
     @property
     def is_valid(self) -> bool:
-        return True
+        # secp224k1 support has been removed from development, but it's stil
+        # available in 3.6 branch.
+        return build_tree.is_mbedtls_3_6()
 
     def arguments(self):
         args = super().arguments()


scripts/mbedtls_framework/ecp.py
```diff
@@ -91,86 +91,6 @@ def arguments(self)-> List[str]:
         args = super().arguments()
         return  ["MBEDTLS_ECP_DP_SECP192R1"] + args
 
-
-class EcpP224R1Raw(bignum_common.ModOperationCommon,
```
gilles-peskine-arm, line 95: We need to keep this for 3.6.

I guess we'll need to skip the 224R1 and 224K1 test cases when called from 1.0. You can check the version of the consuming project with `build_tree.is_mbedtls_3_6()`. I don't know how you arrange to skip a class. There's some documentation in https://mbed-tls.readthedocs.io/en/latest/kb/how-to/add-test-generation/ but it doesn't cover that. The obvious way would be
```
if build_tree.is_mbedtls_3_6():
    class EcpP224R1Raw(…): …
```
but it means `is_mbedtls_3_6()` is checked at load time, and I'm not completely sure if that's ok since the calling script might run `chdir`. The script `generate_ecp_tests.py` has no option to select a different root (only to select a different output directory, which doesn't affect the generated content) so it's probably ok.

valeriosetti: It seems that changing the return value of `is_valid()` those classes from `True` to `False` prevent the data lines from being generated. I think this solves the problem you were mentioning so I will give this a try.

scripts/mbedtls_framework/ecp.py
```diff
@@ -164,7 +165,10 @@ def result(self) -> List[str]:
 
     @property
     def is_valid(self) -> bool:
-        return True
+        if build_tree.is_mbedtls_3_6():
+            return True
+        else:
+            return False
```
gilles-peskine-arm, line 171: That's a bit of a roundabout way to write `return build_tree.is_mbedtls_3_6()`. Not a blocker for me, but it is a blocker for pylint.

gilles-peskine-arm (dismissed):
LGTM, although pylint disapproves

gilles-peskine-arm (approved):
LGTM

felixc-arm (approved):
LGTM :+1:

PR #211, opened 2025-09-16
Fix generate_config_checks.py not actually generating the files

gilles-peskine-arm:
Fix silly mistake in #196 that clogs up the chain of consumers.

## PR checklist

- [x] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/404
- [x] **development PR** provided https://github.com/Mbed-TLS/mbedtls/pull/10340
- [x] **3.6 PR** not required because: new featre

### DIFF
--- a/scripts/mbedtls_framework/config_checks_generator.py
+++ b/scripts/mbedtls_framework/config_checks_generator.py
@@ -218,9 +218,9 @@ def main(branch_data: BranchData) -> None:
                         help='output file location (default: %(default)s)')
     options = parser.parse_args()
     list_only = options.list or options.list_for_cmake
-    output_files = generate_header_files(branch_data,
-                                         options.output_directory,
-                                         list_only=list_only)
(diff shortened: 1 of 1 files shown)


scripts/mbedtls_framework/config_checks_generator.py
```diff
@@ -218,9 +218,9 @@ def main(branch_data: BranchData) -> None:
                         help='output file location (default: %(default)s)')
     options = parser.parse_args()
     list_only = options.list or options.list_for_cmake
-    output_files = generate_header_files(branch_data,
-                                         options.output_directory,
-                                         list_only=list_only)
+    output_files = list(generate_header_files(branch_data,
```
gilles-peskine-arm, line 221: The fact that we all missed this at first shows that this is not a good way to write the code. But right now I'd rather unblock the chain of PR ASAP.

PR #212, opened 2025-09-16
Fix generate_config_checks.py not satisfying make_generated_files --check --root

gilles-peskine-arm:
When the current directory is the project root, have `--list` print relative paths, rather than absolute paths that start with
`build_treeguess_project_root()`.

This is not only nice to humans, but also necessary for `make_generated_files.py --root DIR --check`: it calls this script with `--list` and expects a path that is relative to DIR, not an absolute path that is under the project root.

Fix the build in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/404, hopefully for real this time.

## PR checklist

- [x] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/404
- [x] **development PR** provided https://github.com/Mbed-TLS/mbedtls/pull/10340
- [x] **3.6 PR** not required because: new stuff

### DIFF
--- a/scripts/make_generated_files.py
+++ b/scripts/make_generated_files.py
@@ -239,6 +239,10 @@ def check_generated_files(generation_scripts: List[GenerationScript], root: Path
         for file in generation_script.files:
             file = root / file
             bak_file = file.with_name(file.name + ".bak")
+            if generation_script.optional and not bak_file.exists():
+                # This file is optional and didn't exist before, so
+                # there's nothing to compare to, or clean up.
+                continue
             if not filecmp.cmp(file, bak_file):
                 ref_file = file.with_name(file.name + ".ref")
                 ref_file = root / ref_file
--- a/scripts/make_generated_files.py
+++ b/scripts/make_generated_files.py
@@ -205,6 +205,8 @@ def check_generated_files(generation_scripts: List[GenerationScript], root: Path
     for generation_script in generation_scripts:
         for file in generation_script.files:
             file = root / file
+            if not file.exists():
+                raise Exception(f"Expected generated file does not exist: {file}")
             bak_file = file.with_name(file.name + ".bak")
             if bak_file.exists():
                 bak_file.unlink()
--- a/scripts/make_generated_files.py
+++ b/scripts/make_generated_files.py
@@ -24,7 +24,7 @@ class GenerationScript:
     """
     Representation of a script generating a configuration independent file.
     """
-    # pylint: disable=too-few-public-methods
+    # pylint: disable=too-few-public-methods,too-many-arguments
     def __init__(self, script: Path, files: List[Path],
                  output_dir_option: Optional[str] = None,
                  output_file_option: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM.

gilles-peskine-arm:
This may still not be the final form because `tests/scripts/all.sh -k tf_psa_crypto_check_generated_files` is passing on my machine but failing on https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/404. I'm trying to figure out why. In mbedtls that can be due to insufficient cleanup but in crypto the check is done only in a fresh out-of-source build tree.

gilles-peskine-arm:
> In mbedtls that can be due to insufficient cleanup but in crypto the check is done only in a fresh out-of-source build tree.

No, this is the mbedtls check that's failing. And yes, it is due to insufficient cleanup, I have reproduced it. https://github.com/Mbed-TLS/mbedtls-framework/pull/212/commits/e2c15fbd66ffe5063595d6f2a3ffdbe10124c3a8 isn't doing its intended job.

scripts/make_generated_files.py
```diff
@@ -49,6 +50,12 @@ def __init__(self, script: Path, files: List[Path],
         # positional argument.
         self.output_file_option = output_file_option
 
+        # Optional files are skipped in --check mode if they don't exist.
+        # This normally shouldn't happen, but it can happen during transition
+        # periods where we're adding a new script or a new file, and a
+        # consuming repository hasn't been updated yet.
+        self.optional = optional
```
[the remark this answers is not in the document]
minosgalanakis: I also agree with @mpg that seems like a valid optional input (pun intended) to address a transitoriy edge-case.

scripts/make_generated_files.py
```diff
@@ -205,6 +213,15 @@ def check_generated_files(generation_scripts: List[GenerationScript], root: Path
     for generation_script in generation_scripts:
         for file in generation_script.files:
             file = root / file
+            if not file.exists():
```
minosgalanakis, line 216: Any reason we don't use the same logic as bellow?

`if generation_script.optional and not bak_file.exists():`

scripts/make_generated_files.py
```diff
@@ -222,6 +239,10 @@ def check_generated_files(generation_scripts: List[GenerationScript], root: Path
         for file in generation_script.files:
             file = root / file
             bak_file = file.with_name(file.name + ".bak")
+            if generation_script.optional and not bak_file.exists():
+                # This file is optional and didn't exist before, so
+                # there's nothing to compare to, or clean up.
+                continue
```
minosgalanakis, line 245: Similarly, why are are not throwing an exception as above?

`raise Exception(f"Expected generated file does not exist: {file}")`

minosgalanakis (commented):
looks good, two minor comments.

minosgalanakis (approved):
LGTM

davidhorstmann-arm (approved):
LGTM, thanks!

I've done an incremental review on top of Manuel's and also had a cursory look at the rest of the commits.

PR #223, opened 2025-10-10
tests: migrate tests using secp192[k|r]1 toward secp256[r|k]1 --> PK [1/2]

valeriosetti:
## Description

Prereq for https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/513

## PR checklist

- [x] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/513
- [x] **development PR** not required because: no change
- [x] **3.6 PR** not required because: no backport


data_files/ec_256_prv.pk8.der
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PRIVATE KEY-----
```
mpg: File with `.der` extension looks like PEM.

data_files/ec_256_prv.sec1.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN EC PRIVATE KEY-----
```
mpg: Copy-pasting that into https://asn1js.eu/ and then the content of the existing `ec_256_prv.pem` (in another tab), the two look like they have the exact same structure (but obviously different value for the key material).

OTOH, checking the Makefile, `ec_256_prv.pem` has no rule for generating it (the only file from the `ec_256_prv.*` family that has a rule is `ec_256_prv.comp.pem`). So I'm not sure if we want to use the existing file or generate a new one.

Also, I find the naming confusing, as `ec_256_prv.sec1.pem` doesn't have the same key material as `ec_256_prv.pem` while usually all files that share a common prefix (before the first dot) have the same key material, just different encodings.

I think I'm inclined to:
- not add the `ec_256_prv.sec1` files
- generate `ec_256_prv.der` from the existing `ec_256_prv.pem` (that's the one missing from the existing family)
- add Makefile rules to document (after the fact) how `ec_256_prv` was generated (so, basically the rules you just added, but with the existing names).

Wdyt?

gilles-peskine-arm: I haven't followed all the specifics, but why would we add a file that has exactly the same format as an existing file but different content? Why not either copy the file or align all users on a single name?

data_files/Makefile
```diff
@@ -1839,6 +1839,19 @@ ec_prv.sec1.der: ec_prv.sec1.pem
...
+all_final += ec_256_prv.sec1.der
+
+ec_256_prv.sec1.pem:
+	$(OPENSSL) ec -in $< -out $@ -outform PEM
+all_final += ec_256_prv.sec1.pem
+
+ec_256_prv.sec1.comp.pem:
+	$(OPENSSL) ec -in $< -out $@ -outform PEM
+	$(OPENSSL) ec -in $< -conv_form compressed -outform PEM -out $@
+all_final += ec_256_prv.sec1.pem
```
davidhorstmann-arm: It looks like this was meant to be:
```suggestion
all_final += ec_256_prv.sec1.comp.pem
```

davidhorstmann-arm (changes requested):
One small thing on top of Manuel's points.

Looks good otherwise, thanks!

data_files/Makefile
```diff
@@ -1839,6 +1839,19 @@ ec_prv.sec1.der: ec_prv.sec1.pem
 	$(OPENSSL) pkey -in $< -out $@ -outform DER
 all_final += ec_prv.sec1.der
 
+ec_256_prv.sec1.der:
+	$(OPENSSL) ecparam -name prime256v1 -genkey -outform DER -noout -out $@
+all_final += ec_256_prv.sec1.der
+
+ec_256_prv.sec1.pem:
+	$(OPENSSL) ec -in $< -out $@ -outform PEM
```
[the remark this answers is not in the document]
valeriosetti: Right I run the `openssl` command manually and then I wrote the makefile target :(

data_files/ec_256_prv.sec1.pem
mpg: I'm happy with what's in the makefile, but this file and `ec_256_prv.sec1.comp.pem` are now added to the repo while they're not in the Makefile. Looks like an oversight.

davidhorstmann-arm (approved):
LGTM, thanks!

PR #229, opened 2025-10-22
Update abi_check.py to support standalone tf-psa-crypto

bensze01:
## Description

Move abi_check.py to the framework, and update it to support a standalone tf-psa-crypto

## PR checklist

- [ ] **framework PR**: https://github.com/Mbed-TLS/mbedtls-framework/pull/229
- [ ] **mbedtls-test PR**: https://github.com/Mbed-TLS/mbedtls-test/pull/241
- [ ] **TF-PSA-Crypto PR**: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/541
- [ ] **development PR**: https://github.com/Mbed-TLS/mbedtls/pull/10464
- [ ] **3.6 PR**: https://github.com/Mbed-TLS/mbedtls/pull/10465

### DIFF
--- a/scripts/check-python-files.sh
+++ b/scripts/check-python-files.sh
@@ -74,7 +74,7 @@ $PYTHON -m mypy framework/scripts/*.py framework/scripts/mbedtls_framework/*.py 
 }
 
 $PYTHON -m mypy scripts/*.py tests/scripts/*.py || {
-    echo >&2 "pylint reported errors in the parent repository"
+    echo >&2 "mypy reported errors in the parent repository"
     ret=1
 }
 


scripts/mbedtls_framework/abi_check.py
```diff
@@ -151,6 +151,7 @@ def _setup_logger(self):
 
     @staticmethod
     def check_abi_tools_are_installed():
+        return
```
gilles-peskine-arm: Stray debugging bypass.

bensze01:
2 more things that might be good to check:
- I tried to minimize the history of Mbed TLS we needed to pull in, so https://github.com/Mbed-TLS/mbedtls-framework/commit/6a9be3de1f2b4b9ba16b456fd5d4bedd181ecbcd is the  merge commit of the most recent PR that modified `abi_check.py` (Mbed-tls/mbedtls#10416)
- [`49e9d30`](https://github.com/Mbed-TLS/mbedtls-framework/tree/49e9d30bcd8b59af0402a0f8df11b1edca9822a1) deletes all files from the mbedtls tree except the ones to be moved (`abi_check.py`)

scripts/check-python-files.sh
```diff
@@ -55,7 +55,13 @@ elif [ "$1" = "--can-mypy" ]; then
...
 
 echo 'Running pylint ...'
-$PYTHON -m pylint framework/scripts/*.py framework/scripts/mbedtls_framework/*.py scripts/*.py tests/scripts/*.py || {
+# Temporary workaround while moving the bulk of abi_check.py to the framework
+# Check abi_check.py separately from the rest of the files, so it's not flagged
+# for code duplication.
+find framework/scripts/*.py framework/scripts/mbedtls_framework/*.py scripts/*.py tests/scripts/*.py \
+     -path scripts/abi_check.py \
+        -exec $PYTHON -m pylint {} + \
+     -o -exec $PYTHON -m pylint {} + || {
```
gilles-peskine-arm: That doesn't work: `find` ignores the exit status of `-exec` commands. It's also rather obscure.

We could ask git to list the files. That's what we do in most check scripts. It means that generated files aren't checked (but we don't expect to ever generate a python script) and files that a developer might have in their working copy won't be checked (which I think is a good thing).

```
# Run pylint on both this repository and the framework in a single invocation.
# It has to be a single invocation because we want to be told if there is
# code duplication between the main repository and the framework.
# We don't check other submodules because we're ok (for now) with code
# duplication between mbedtls and tf-psa-crypto, and also because we don't
# want to check "foreign" submodules that we might add such as mldsa-native.
$PYTHON -m pylint $(
    # Temporarily exclude the local abi_check.py while it is being moved to
    # the framework.
    git ls-files '*.py' ':!docs/*' ':!tests/psa-client-server/psasim/*' \
        ':!scripts/abi_check.py'
    # Exclude the code_wrapper subdirectory until we fix
    # https://github.com/Mbed-TLS/mbedtls-framework/issues/259
    git -C framework ls-files '*.py' ':!psasim/*' \
            ':!scripts/mbedtls_framework/code_wrapper/*' |
        sed 's!^!framework/!'
) || {
    …
}
```

(I'm deliberately avoiding `xargs` because it could end up silently running in multiple chunks. That could also happen with `find -exec +`, by the way.)

gilles-peskine-arm: Or we could keep it simple and let pylint do the file enumeration. But I can't seem to get `pylint --ignore` to work.

PR #238, opened 2025-11-17
Move psasim into framework

minosgalanakis:
## Description

Part of the work required by Mbed-TLS/mbedtls#10487. Psasim was originally placed in framework directory, but moved to mbedtls while we were workign on it. Now this PR is migrating it back, while preserving the history

- [x] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/678
- [x] **development PR** provided  Mbed-TLS/mbedtls#10511
- [x] **3.6 PR** provided # | not required because: psasim is not part of 3.6 LTS

### DIFF
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -267,7 +267,18 @@ pre_initialize_variables () {
     # defined in this script whose name starts with "component_".
     ALL_COMPONENTS=$(compgen -A function component_ | sed 's/component_//')
 
-    PSASIM_PATH='framework/psasim/'
+
+    # Allow overriding PSASIM_PATH with the following priority:
+    #  1) Environment override
+    #  2) Mbed TLS's version of psa-client-server
+    #  3) Fallback to the framework location (default after move)
+    if [ -z "${PSASIM_PATH+set}" ]; then
+        if [ -d tests/psa-client-server ]; then
+            PSASIM_PATH='tests/psa-client-server/psasim/'
+        else
+            PSASIM_PATH='framework/psasim/'
+        fi
+    fi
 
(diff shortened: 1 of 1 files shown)


minosgalanakis:
This PR fails the CI because of https://github.com/Mbed-TLS/mbedtls-framework/pull/238/commits/cbf166ee63e8a4a516fa6f901325a2111013295d. We know the code works because it has been tested as part of the https://github.com/Mbed-TLS/mbedtls/pull/10511 

The issue is a chicken and an egg situation, where this PR needs to be be merged without https://github.com/Mbed-TLS/mbedtls-framework/pull/238/commits/cbf166ee63e8a4a516fa6f901325a2111013295d and at that point will have two instances of PSA-sim. One in `tests/psa-client-server/psasim/` and one in 'framework/psasim/'

Then I need to come up with acceptable merge strategy because there is a circular dependency between the mbedtls side using the new location and https://github.com/Mbed-TLS/mbedtls-framework/pull/238/commits/cbf166ee63e8a4a516fa6f901325a2111013295d which needs to be in the framework.

The first part of the PR can be reviewed for content though.

psasim/include/client.h
```diff
@@ -14,7 +14,10 @@ extern "C" {
 
 #include <stdint.h>
 #include <stddef.h>
-#include <psa/error.h>
+
```
gilles-peskine-arm: Please use [`mbedtls-move-to-framework‎`](https://github.com/Mbed-TLS/mbedtls-docs/pull/145) instead of importing a rewritten history. With a rewritten history, we don't get the same commits.

gilles-peskine-arm (changes requested):
Please use the proper way to transfer the history.

PR #243, opened 2025-12-01
Add support for TF-PSA-Crypto test driver

ronald-cron-arm:
## Description
Add support for driver dispatch testing in TF-PSA-Crypto

## PR checklist
- [x] **TF-PSA-Crypto PR** provided Mbed-TLS/TF-PSA-Crypto#586 
- [x] **development PR** not required because: TF-PSA-Crypto only work
- [x] **3.6 PR** not required because: TF-PSA-Crypto only work

### DIFF
--- a/scripts/mbedtls_framework/test_driver.py
+++ b/scripts/mbedtls_framework/test_driver.py
@@ -167,7 +167,7 @@ class TestDriverGenerator:
                   self.__get_dst_relpath(file.relative_to(self.src_dir))
             dst.parent.mkdir(parents=True, exist_ok=True)
             self.__write_test_driver_file(file, dst, headers,\
-                                          identifiers_to_prefix, self.driver)
+                                          identifiers_to_prefix)
 
     @staticmethod
     def __get_code_files(root: Path) -> List[Path]:
@@ -247,11 +247,9 @@ class TestDriverGenerator:
 
         return identifiers
 
-    @staticmethod
-    def __write_test_driver_file(src: Path, dst: Path,
+    def __write_test_driver_file(self, src: Path, dst: Path,
                                  headers: Set[str],
-                                 identifiers_to_prefix: Set[str],
-                                 driver: str) -> None:
+                                 identifiers_to_prefix: Set[str]) -> None:
         """
         Write a test driver file to `dst` based on the contents of `src` with
         two transformations: rewriting of `#include` directives and identifier
@@ -260,7 +258,7 @@ class TestDriverGenerator:
         1. Rewrite header inclusions
            Any `#include` directive whose header basename matches an entry of
            `headers` is rewritten so that the basename is prefixed with
-           `{driver}-`. Directory components (if any) are preserved.
+           `{self.driver}-`. Directory components (if any) are preserved.
 
            Example:
                #include "mbedtls/private/aes.h"
@@ -289,9 +287,10 @@ class TestDriverGenerator:
                code base.
 
         2. Rename selected identifiers
-           Each identifier in `identifiers_to_prefix` is prefixed with `driver`.
-           Case is preserved: if the identifier is all-uppercase, then the
-           uppercase form of `driver` is used, the lowercase form otherwise.
+           Each identifier in `identifiers_to_prefix` is prefixed with
+           `self.driver`. Case is preserved: if the identifier is all-uppercase,
+           then the uppercase form of `driver` is used, the lowercase form
+           otherwise.
 
            Examples:
                `MBEDTLS_AES_C` becomes `LIBTESTDRIVER1_MBEDTLS_AES_C`
@@ -308,12 +307,8 @@ class TestDriverGenerator:
                 Basenames of headers whose includes should be rewritten.
 
             identifiers_to_prefix (Set[str]):
-                Identifiers that must be renamed by prefixing with `driver`
+                Identifiers that must be renamed by prefixing with `self.driver`
                 (using uppercase or lowercase depending on the identifier's casing).
-
-            driver (str):
-                The name of the driver. Used as a prefix for rewritten include
-                paths and identifier names.
         """
         text = src.read_text(encoding="utf-8")
 
@@ -324,14 +319,14 @@ class TestDriverGenerator:
         def repl_header_inclusion(m: Match) -> str:
             parts = m.group(2).split("/")
             if parts[-1] in headers:
-                path = "/".join(parts[:-1] + [driver + "-" + parts[-1]])
+                path = "/".join(parts[:-1] + [self.driver + "-" + parts[-1]])
                 return f'{m.group(1)}{path}{m.group(3)}'
             return m.group(0)
         intermediate_text = include_line_re.sub(repl_header_inclusion, text)
 
         c_identifier_re = re.compile(r"\b[A-Za-z_][A-Za-z0-9_]*\b")
-        prefix_uppercased = driver.upper()
-        prefix_lowercased = driver.lower()
+        prefix_uppercased = self.driver.upper()
+        prefix_lowercased = self.driver.lower()
 
         def repl(m: Match) -> str:
             identifier = m.group(0)


tests/src/drivers/test_driver_aead.c
```diff
@@ -15,13 +15,16 @@
 
 #include "mbedtls/constant_time.h"
 
-#if defined(MBEDTLS_TEST_LIBTESTDRIVER1)
+#if defined(TF_PSA_CRYPTO_TEST_LIBTESTDRIVER1)
+typedef psa_key_attributes_t libtestdriver1_psa_key_attributes_t;
```
gilles-peskine-arm: Since this is repeated, it feels like it should be in a header somewhere, probably `tests/include/test/drivers/test_driver_common.h`.

tests/src/drivers/test_driver_aead.c
```diff
@@ -15,13 +15,16 @@
...
+#include "../../libtestdriver1/src/psa_crypto_aead.h"
```
gilles-peskine-arm: It took me a while to understand this path. It's weird to go up two levels. My first thought was that the include path was relative to the file that this directive is in, but that didn't work, and a path relative to the framework would be fragile since it could be different in in-tree vs out-of-tree builds. Then I noticed the commit message

> The paths are relative to `drivers/libtestdriver1/include`.

Ok, that makes sense, but then why not `../src/psa_crypto_aead.h`? But that wouldn't necessarily work since it could pick up the one under `/drivers/builtin`. Ok, then it does have to be what you wrote, although I would prefer
```
#include <../../libtestdriver1/src/psa_crypto_aead.h>
```
since we know this isn't relative to the location of the file, it comes from the include path.

And I would like this explained in a comment.

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Only "*.h" and "*.c" files are copied. Files whose names match any of the
+        patterns in `exclude_files` are excluded.
+
+        The subdirectory inside `include` is renamed to `driver` in the test driver
+        tree, and header inclusions are adjusted accordingly.
+        """
+        include = src_dir / "include"
+        if not include.is_dir():
+            raise RuntimeError(f'Do not find "include" directory in {src_dir}')
```
gilles-peskine-arm: Minor: “do not find” is an instruction. A complaint should be “Did not find” or more usually “… not found”.

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,219 @@
...
+        Rewrite `#include` directives in `file` that refer to `src_include_dir/...`
+        so that they instead refer to `driver/...`.
+
+        For example:
+            #include "mbedtls/private/aes.h"
+        becomes:
+            #include "libtestdriver1/private/aes.h"
+        """
+        include_line_re = re.compile(
+            fr'^\s*#\s*include\s*([<"])\s*{src_include_dir}/([^>"]+)\s*([>"])',
```
gilles-peskine-arm: You can't have spaces inside the quotes/brackets.

```suggestion
            fr'^\s*#\s*include\s*([<"]){src_include_dir}/([^>"]+)([>"])',
```

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,219 @@
...
+        def repl(m: Match) -> str:
+            identifier = m.group(0)
+            if identifier in identifiers:
+                if identifier[0].isupper():
+                    return f"{prefix_uppercased}_{identifier}"
+                else:
+                    return f"{prefix_lowercased}_{identifier}"
+            return identifier
+
+        new_text = c_identifier_re.sub(repl, text)
```
gilles-peskine-arm: You can make regexps do the work:

```suggestion
        lower_identifier_re = re.compile(r'\b(' + '|'.join(ident for ident in identifiers if ident[0].islower()) + r')\b')
        upper_identifier_re = re.compile(r'\b(' + '|'.join(ident for ident in identifiers if ident[0].isupper()) + r')\b')
        text = file.read_text(encoding="utf-8")
        prefix_uppercased = prefix.upper()
        prefix_lowercased = prefix.lower()
        new_text = lower_identifier_re.sub(prefix.lower() + r'_\1',
                                           upper_identifier_re.sub(prefix.upper() + r'_\1', text))
```

gilles-peskine-arm (changes requested):
I've fully reviewed 61f15a6baee21fb9a44c0471f2a749302942de3e, [as well as the Python code](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/586#pullrequestreview-3536355545) in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/586.

I like the methodology, but the Python code sometimes feels hard to follow compared to what it does.

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,222 @@
...
+
+        if exclude_files is None:
+            exclude_files = set()
+
+        for file in iter_code_files(src_dir):
+            if any(fnmatch(file.name, pattern) for pattern in exclude_files):
+                continue
+            dst = self.dst_dir / file.relative_to(src_dir)
+            dst.parent.mkdir(parents=True, exist_ok=True)
+            shutil.copy2(file, dst)
```
minosgalanakis: I would keep and return a list of all the files that have been created

minosgalanakis (commented):
Functionally works, but there is some work required for the design

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,293 @@
...
+        excluding the files whose basename match any of the patterns in
+        `self.exclude_files`.
+        """
+        out = []
+        for file in self.__get_code_files(self.src_dir):
+            if not any(fnmatch(file.name, pattern) for pattern in self.exclude_files):
+                out.append(file)
+        return out
+
+    def __get_dst_relpath(self, src_relpath: Path) -> Path:
```
minosgalanakis: **Not a blocker** but this could also be a staticmethod, and `self.driver` to be provided as a prefix argument.

minosgalanakis (commented):
Looks quite good, some minor comments only, but I wouldn't block the PR if the second reviewer is happy with it.

The python code is clean and easy to understand and maintain

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,293 @@
...
+                are correct.
+
+                The source directory is expected to contain:
+                - an `include` directory
+                - an `src` directory
+                - the `include/` directory contains exactly one subdirectory
+
+            dst_dir (Path):
+                Path to the destination directory where the rewritten tree will
+                be created.
```
gilles-peskine-arm: It must exist.

Which, again, I find strange: why not create it?

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,293 @@
...
+        - struct tags (s)
+        - typedefs (t)
+        - union tags (u)
+        - global variables (v)
+
+        This method requires `Universal Ctags`. The command used here has been
+        validated with Universal Ctags 5.9.0 (the default `ctags` on Ubuntu
+        22.04 and 24.04). `Exuberant Ctags` is not compatible: it does not
+        support all `--c-kinds` flags used here, and will either fail with an
+        error or produce incomplete results.
```
gilles-peskine-arm: What's the problem with Exuberant? I tried running `tests/scripts/generate_test_driver.py` in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/586 as of 2b70a5e35e5b5b92aa24295cfbaf166db61160e8 and got identical results with Exuberant Ctags 5.8 (compiled manually), Exuberant Ctags 5.9~svn20110310 (Ubuntu 24.04), Universal Ctags 5.9.0 (Ubuntu 24.04) and Universal Ctags 6.2.0 (Ubuntu 26.04).

If it's possible to get incomplete results, the script should check the ctags implementation and error out. Possibly by calling `ctags-universal`, which is the command for Universal Ctags on Debian and derivatives, but not on other platforms (e.g. Brew on macOS just installs `ctags` and you can only have one of Exuberant or Universal, and you might have to fiddle with `$PATH` to put Brew ahead of the system `ctags` which would error out on invocation).

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,219 @@
...
+
+        def repl(m: Match) -> str:
+            nonlocal changed
+            header = m.group(2)
+            if header in headers:
+                changed = True
+                return f'#include {m.group(1)}{driver}/{header}{m.group(3)}'
+            return m.group(0)
+
+        new_text = include_line_re.sub(repl, text)
```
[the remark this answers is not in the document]
gilles-peskine-arm: Thanks for trying it out. I've done this in similar situations, but I might have had significantly fewer items.

gilles-peskine-arm (changes requested):
Mostly LGTM, but I want to clarify the ctags implementation/version dependency. The comments I flagged as “minor” are optional.

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def __get_src_code_files(self) -> List[Path]:
+        """
+        Return all "*.c" and "*.h" files found recursively under the
+        `include` and `src` subdirectories of the source directory `self.src_dir`
+        excluding the files whose basename match any of the patterns in
+        `self.exclude_files`.
+        """
+        out = []
+        for file in self.__get_code_files(self.src_dir):
```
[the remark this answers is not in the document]
ronald-cron-arm: I haven't changed that.

scripts/mbedtls_framework/test_driver.py
ronald-cron-arm: This one is only used in instance context thus I prefer to keep like this. I've aligned "__write_test_driver_files" with that.

scripts/mbedtls_framework/test_driver.py
```diff
@@ -0,0 +1,293 @@
...
+            if not any(fnmatch(file.name, pattern) for pattern in self.exclude_files):
+                out.append(file)
+        return out
```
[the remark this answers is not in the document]
ronald-cron-arm: I haven't taken that one.

tests/include/test/drivers/test_driver_common.h
```diff
@@ -25,4 +25,66 @@
...
+ *
+ * Uncrustify is not happy with the macros, temporarily disable it.
+ *
+ * *INDENT-OFF*
+ */
+#if defined(TF_PSA_CRYPTO_TEST_LIBTESTDRIVER1)
+#define LIBTESTDRIVER1_PSA_DRIVER_INTERNAL_HEADER(basename) \
+    <../src/libtestdriver1-basename>
+#else
+#if MBEDTLS_VERSION_MAJOR < 4
```
gilles-peskine-arm: Minor: could be `#elif`

tests/include/test/drivers/test_driver_common.h
```diff
@@ -66,11 +66,7 @@ typedef psa_crypto_driver_pake_step_t libtestdriver1_psa_crypto_driver_pake_step
  * - In TF-PSA-Crypto, the libtestdriver1 library code is located in
  *   `drivers/libtestdriver1`. The header path is relative to
  *   `drivers/libtestdriver1/include` and has the form:
- *     `../../libtestdriver1/src/xyz`
- *
- *   Note: the path cannot simply be `../src/xyz`, because that could refer to
- *   a header under `drivers/builtin/src`. The directory `drivers/builtin/include`
- *   is also in the header search path when compiling test drivers.
```
gilles-peskine-arm: I lost track of why “the path cannot simply be `../src/xyz`” no longer applies. It's because the headers are now renamed with a prefix, e.g. `<mbedtls/private/libtestdriver1-aes.h>`, whereas in the original design they were just copied to e.g. `<mbedtls/private/aes.h>` so you had to make sure not to land on the one under `builtin`. This is only relevant to the history within this PR, so it doesn't warrant a comment in the file, but I'm leaving it in writing in case we have doubts later.

PR #257, opened 2025-12-17
[framework] tests: pk: add a common function to create a PSA key out of predefined keys

valeriosetti:
## Description

Prerequisite for https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/600

## PR checklist

- [ ] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/600
- [ ] **development PR** not required because: no change there
- [ ] **3.6 PR** not required because: no backport (at least not for now)

### DIFF
--- a/tests/include/test/pk_helpers.h
+++ b/tests/include/test/pk_helpers.h
@@ -50,7 +50,7 @@ typedef enum {
  *                   key is not the in list of known ones.
  */
 int mbedtls_pk_helpers_get_predefined_key_data(psa_key_type_t key_type, psa_key_bits_t key_bits,
-                                                const uint8_t **output, size_t *output_len);
+                                               const uint8_t **output, size_t *output_len);
 
 /**
  * Create a PSA key using prefined key data.
--- a/tests/src/pk_helpers.c
+++ b/tests/src/pk_helpers.c
@@ -64,7 +64,7 @@ static mbedtls_pk_helpers_predefined_key_t predefined_keys_psa[] = {
 };
 
 int mbedtls_pk_helpers_get_predefined_key_data(psa_key_type_t key_type, psa_key_bits_t key_bits,
-                                                const uint8_t **output, size_t *output_len)
+                                               const uint8_t **output, size_t *output_len)
 {
     for (size_t i = 0; i < ARRAY_LENGTH(predefined_keys_psa); i++) {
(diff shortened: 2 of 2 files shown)


tests/include/test/pk_helpers.h
```diff
@@ -0,0 +1,38 @@
...
+#if defined(MBEDTLS_PK_C)
+
+#include <psa/crypto.h>
+#include <mbedtls/pk.h>
+
+typedef enum {
+    TEST_PK_WRAP_PSA,
+    TEST_PK_COPY_FROM_PSA,
+    TEST_PK_COPY_PUBLIC_FROM_PSA,
+} pk_context_populate_method_t;
```
gilles-peskine-arm: Why isn't parsing listed here?

tests/include/test/pk_helpers.h
```diff
@@ -0,0 +1,38 @@
+/*
+ * Helper functions for PK
```
gilles-peskine-arm: Please document that this is only for TF-PSA-Crypto 1.0 and above.

tests/src/pk_helpers.c
```diff
@@ -103,26 +106,31 @@ mbedtls_svc_key_id_t mbedtls_pk_helpers_make_psa_key_from_predefined(psa_key_typ
...
-void mbedtls_pk_helpers_populate_context(mbedtls_pk_context *pk, mbedtls_svc_key_id_t key_id,
+int mbedtls_pk_helpers_populate_context(mbedtls_pk_context *pk, mbedtls_svc_key_id_t key_id,
                                          pk_context_populate_method_t method)
 {
+    int ret;
+
     switch (method) {
         case TEST_PK_WRAP_PSA:
-            TEST_EQUAL(mbedtls_pk_wrap_psa(pk, key_id), 0);
+            ret = mbedtls_pk_wrap_psa(pk, key_id);
```
gilles-peskine-arm: It would be more informative to keep `TEST_EQUAL(mbedtls_pk_xxx(), 0)`, then the error message would show which function was called.

What I usually do in such auxiliary functions is:

```
int aux_func(...) {
    int ok = 0;
    TEST_ASSERT(...);
    ...
    ok = 1;
exit:
    return ok;
}
```

and then the caller does
```
if (!aux_func(...)) goto exit;
```

PR #261, opened 2026-01-05
Fix --quiet (all.sh, check_names.py)

mpg:
## Description

I like to run `all.sh -q <some_check_components>` as part of my pre-push script, but these days it's not really as quiet as advertised ("Only output component names, and errors if any."). This PR addresses most of that:

1. Stop printing make invocations (to stderr).
2. Address error message about root Makefile in development -> here it's not just about silencing the error, something was actually wrong.
3. Make the `-q` option of `check_names` more effective - also make normal output go to stdout not stderr.

## PR checklist

- [x] **TF-PSA-Crypto PR** not required because: all affected scripts in the frameworks
- [x] **development PR** not required because: all affected scripts in the frameworks
- [x] **3.6 PR** not required because: all affected scripts in the frameworks

### DIFF
--- a/scripts/check_names.py
+++ b/scripts/check_names.py
@@ -51,6 +51,16 @@ import project_scripts # pylint: disable=unused-import
 from mbedtls_framework import build_tree
 
 
+class MaxLevelFilter(logging.Filter):
+    """Allow records with level <= max_level."""
+    def __init__(self, max_level):
+        super().__init__()
+        self.max_level = max_level
+
+    def filter(self, record):
+        return record.levelno <= self.max_level
+
+
 # Naming patterns to check against. These are defined outside the NameCheck
 # class for ease of modification.
 PUBLIC_MACRO_PATTERN = r"^(MBEDTLS|PSA|TF_PSA_CRYPTO)_[0-9A-Z_]*[0-9A-Z]$"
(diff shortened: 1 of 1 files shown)


scripts/check_names.py
```diff
@@ -791,11 +801,13 @@ def parse_symbols(self):
             source_dir = os.getcwd()
             build_dir = tempfile.mkdtemp()
             os.chdir(build_dir)
-            subprocess.run(
+            result = subprocess.run(
```
gilles-peskine-arm: Partly preexisting: it would be more robust to use `subprocess.check_output` (formerly `check_call`).

scripts/check_names.py
```diff
@@ -1178,15 +1192,28 @@ def main():
...
 
     # Configure the global logger, which is then passed to the classes below
     log = logging.getLogger()
-    log.setLevel(logging.DEBUG if args.verbose else logging.INFO)
-    log.addHandler(logging.StreamHandler())
+    log.setLevel(logging.DEBUG if args.verbose else
+                 logging.INFO if not args.quiet else
+                 logging.WARN)
+
+    # stdout handler: DEBUG/INFO only
```
gilles-peskine-arm: Why not send everything to stderr? That's the normal place for informational messages, as opposed to application data which goes to stdout. Separating two logging streams also creates the risk that they'll end up displayed out of order due to buffering (which I think doesn't happen on Jenkins, it did happen on Travis).

PR #268, opened 2026-01-09
all.sh: fail if a file is missing on a "not grep ..." line

### LINKED ISSUE #266: all.sh: improve fragile `not grep` idiom
We often check for the absence of a function in a file with the idiom `not grep REGEXP FILE[...]` in `all.sh`. The function `not` is a wrapper around `!` that takes care of failing the component if the command exits with the status 0. As noted in https://github.com/Mbed-TLS/mbedtls/pull/10517#discussion_r2675303189 this is fragile: it'll pass if the file doesn't exist.

The goal of this issue is:

1. Pick a name for a “not grep” function. `not_grep` is natural but might be a bit misleading since it's not just `! grep`.
2. Define a new function that does what we want. I believe the semantics is: `not_grep REGEXP FILE[...]` passes if all the files exist and the regexp is not found in any of the files. Note that the files can be expressed as a pattern, and since we use bash with default settings, if the pattern doesn't match any file, it'll remain unexpanded and the function will see it as a file name that doesn't exist.
3. Systematically replace `not grep` by the new function in `tests/scripts/components-*.sh` in all branches.

gilles-peskine-arm:
Fixes https://github.com/Mbed-TLS/mbedtls-framework/issues/266

This is transparent since `not grep` automatically redirects to the new function. I'm not sure whether this should be transitional or if we should keep this automation.

Needs preceding PR: https://github.com/Mbed-TLS/mbedtls/pull/10559, https://github.com/Mbed-TLS/mbedtls/pull/10562, TODO to fix the problems that this reveals.

## PR checklist

- [x] **TF-PSA-Crypto PR** provided # | not required because: transparent change
- [x] **development PR** provided # | not required because: transparent change
- [x] **3.6 PR** provided # | not required because: transparent change

### DIFF
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -810,10 +810,7 @@ pre_setup_keep_going () {
 ## to be failed.
 not_grep () {
     declare ret=0
-    # Add /dev/null as a file name. This forces the file name to be displayed
-    # if there is a single file name, and prevents blocking waiting on stdin
-    # if this function is accidentally called with zero file names.
-    grep "$@" /dev/null || ret=$?
+    grep -H "$@" || ret=$?
     if [[ $ret -eq 0 ]]; then
         # A match was found, and displayed on stdout.
         # This is a test failure.
--- a/scripts/all-core.sh
+++ b/scripts/all-core.sh
@@ -740,6 +740,7 @@ pre_setup_keep_going () {
             *test*) true;; # make test, tests/stuff, env V=v tests/stuff, ...
             *make*check*) true;;
             "grep "*) true;;
+            "not_grep "*) true;;
             "[ "*) true;;
             "! "*) true;;
             *) false;;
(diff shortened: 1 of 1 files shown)


mpg (dismissed):
LGTM, thanks!

mpg (commented):
LGTM but the CI does not seem to agree.

scripts/all-core.sh
```diff
@@ -791,9 +781,45 @@ pre_setup_keep_going () {
     }
 }
 
+## not_grep [GREP_OPTION...] REGEXP FILE[...]
+## Assert that none of the files contains a match for REGEXP.
+## Errors such as a missing file are also assertion failures.
+## If the assertion fails, print an error and consider the test case
+## to be failed.
+not_grep () {
+    #set -x
```
valeriosetti: this seems like a comment to me. If so, can we remove it?

scripts/all-core.sh
```diff
@@ -791,9 +781,45 @@ pre_setup_keep_going () {
...
+    elif [[ $ret -ne 1 ]]; then
+        # If grep exited with error code 1, grep ran find and found that
+        # there were no matches, which means the assertion is true.
+        # If grep exited with an error code >=2, then something went
+        # wrong, probably a missing file (or e.g. a command syntax error).
+        # grep already printed an error message, so we just need to assert
+        # the condition and the ERR trap will mark the test case as failed.
+        # But make sure that the failure report indicates the grep command.
+        report_failed_command="not_grep $*"
+        (exit $ret)
```
valeriosetti, line 829: Question for my limited knowledge of bash: this trick instead of `false` is because we want to return the error code given by `grep`, right?

valeriosetti (changes requested):

scripts/all-core.sh
gilles-peskine-arm: I've rebased on top of `main` because an old branch wouldn't pass the CI. While I was at it, I amended the commit that added this spurious line.

scripts/all-core.sh
gilles-peskine-arm: That's right. There's no standard shell command that exits itself with a specified status, only the `exit` command which exits the shell. So to get the effect of a command exiting with a given status, I use the `exit` builtin in a subshell.

valeriosetti (approved):

PR #282, opened 2026-02-25
Generate MLDSA test cases for the driver and dispatch layers

gilles-peskine-arm:
Framework support for https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/700.

Continues https://github.com/Mbed-TLS/mbedtls-framework/pull/278.

Needs preceding PR:

* [x] https://github.com/Mbed-TLS/mbedtls-framework/pull/278
* [x] https://github.com/Mbed-TLS/mbedtls-framework/pull/283

## PR checklist

- [x] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/700
- [x] **development PR** provided # | not required because: crypto only
- [x] **3.6 PR** provided # | not required because: new feature

### DIFF
--- a/util/mbedtls_maintainer/__init__.py
+++ b/util/mbedtls_maintainer/__init__.py
@@ -1,0 +1,3 @@
+# This file needs to exist to make mbedtls_maintainer a package.
+# Among other things, this allows modules in this directory to make
+# relative imports.
--- a/util/mbedtls_maintainer/mldsa_test_generator.py
+++ b/util/mbedtls_maintainer/mldsa_test_generator.py
@@ -82,45 +82,6 @@ class API:
     def secret_is_seed(cls) -> bool:
         return True
 
-
-class PQCPAPI(API):
-    """Test mldsa-native entry points."""
-
-    @classmethod
-    def function(cls, func: str, kl: int) -> str:
-        return f'{func}_{kl}'
-
-    @classmethod
-    def metadata_arguments(cls,
-                           _kl: int,
-                           _pair: bool,
-                           _deterministic: Optional[bool]) -> List[str]:
-        return []
-
-    @classmethod
-    def secret_is_seed(cls) -> bool:
-        return False
-
-
-def one_mldsa_key_pair_from_seed(key: Key,
-                                 descr: str) -> test_case.TestCase:
-    """Construct one test case for mldsa-native keypair_internal()."""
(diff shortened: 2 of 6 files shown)


util/mbedtls_maintainer/__init__.py
bjwtaylor, line 1: Does this file need the standard copyright header?

util/mbedtls_maintainer/mldsa_test_generator.py
```diff
@@ -0,0 +1,259 @@
...
+        ] + self.final_arguments())
+        tc.set_description(f'MLDSA-{key.kl} export public key from seed {descr}')
+        return tc
+
+    def one_mldsa_sign_deterministic_pure(self,
+                                          key: Key,
+                                          message: bytes,
+                                          descr: str) -> test_case.TestCase:
+        """Construct one test case for deterministic signature."""
+        signature = key.sign_message(message, deterministic=True)
```
bjwtaylor, line 103: Are we 100% sure that we will never support anything but deterministic? Id not would this be better as an input variable? Other PR's seem to indicate maybe this is temporary?

bjwtaylor (commented):
Still need to do a full review, however some initial comments. I also think it needs a rebase as currently with the upstream versions the build produces fatal error: test/fork_helpers.h: No such file or directory, though this is a minor point.

util/mbedtls_maintainer/__init__.py
gilles-peskine-arm: It could, but there's nothing in the file that would actually get copyright protection. (Then again, there have been copyright claims on the empty program…) I copied an existing file.

util/mbedtls_maintainer/mldsa_test_generator.py
gilles-peskine-arm: We are 100% sure that we will support non-deterministic signatures. But they're not implemented yet. And this method is only for deterministic signatures. Non-deterministic signature testing will have a different method because the test data looks very different.

valeriosetti (approved):

gilles-peskine-arm:
The last CI run failed because the branch is out of date. I've pushed a merge commit with the head of `main`. This doesn't invalidate the existing approval.

bjwtaylor (approved):
Approve subject to CI

PR #292, opened 2026-03-26
Add outcome analysis to TF-PSA-Crypto: framework support

gilles-peskine-arm:
Support for https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/729

* In coverage analysis, distinguish between _uncovered_ tests (which must not be executed) and _ignored_ tests (which may or may not be executed).
* During a transition period, consuming branches can keep defining `IGNORED_TESTS` in the coverage task. These tests were formerly enforced not-executed, but are now ignored.
* Once consuming branches update their framework pointer, they will rename their `IGNORED_TESTS` definition to `UNCOVERED_TESTS`. This will restore the old semantics. The consuming PR in the checklist below do this.

## PR checklist

- [x] **development PR** https://github.com/Mbed-TLS/mbedtls/pull/10676
- [x] **TF-PSA-Crypto development PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/729
- [ ] **TF-PSA-Crypto 1.1 PR** TODO
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/292
- [ ] **4.1 PR** TODO
- [x] **3.6 PR** https://github.com/Mbed-TLS/mbedtls/pull/10677

### DIFF
--- a/scripts/mbedtls_framework/outcome_analysis.py
+++ b/scripts/mbedtls_framework/outcome_analysis.py
@@ -229,7 +229,7 @@ class CoverageTask(Task):
 
     # Test cases whose suite and description are matched by an entry in
     # UNCOVERED_TESTS are expected to be never executed.
-    # Tests matched by IGNORED_TESTS are ignored entierly.
+    # Tests matched by IGNORED_TESTS are ignored entirely.
     # All other test cases are expected to be executed at least once.
 
     UNCOVERED_TESTS: TestCaseSetDescription = {}
@@ -313,9 +313,10 @@ class DriverVSReference(Task):
     DRIVER = ''
     # Ignored test suites (without the test_suite_ prefix).
     IGNORED_SUITES = [] #type: typing.Sequence[str]
-    # Ignored test cases. Despite the name, these test case are not
-    # completely ignored: they must be skipped by drivers, indicating
-    # a spurious entry.
+    # Ignored test cases. Despite the name, these test cases are not
+    # completely ignored: they must be skipped by driver tests. If they
+    # are not skipped, this indicates a spurious entry and the analysis will
+    # complain.
     IGNORED_TESTS: TestCaseSetDescription = {}
 
     def __init__(self, options) -> None:
--- a/scripts/mbedtls_framework/outcome_analysis.py
+++ b/scripts/mbedtls_framework/outcome_analysis.py
@@ -220,19 +220,29 @@ class CoverageTask(Task):
     """Analyze test coverage."""
 
     # Test cases whose suite and description are matched by an entry in
-    # IGNORED_TESTS are expected to be never executed.
+    # UNCOVERED_TESTS are expected to be never executed.
+    # Tests matched by IGNORED_TESTS are ignored entierly.
     # All other test cases are expected to be executed at least once.
 
+    UNCOVERED_TESTS: TestCaseSetDescription = {}
     IGNORED_TESTS: TestCaseSetDescription = {}
 
     def __init__(self, options) -> None:
         super().__init__(options)
         self.full_coverage = options.full_coverage #type: bool
+        self.uncovered_tests = TestCaseSet(self.UNCOVERED_TESTS)
         self.ignored_tests = TestCaseSet(self.IGNORED_TESTS)
 
     @staticmethod
     def section_name() -> str:
         return "Analyze coverage"
+
+    def note_ignored_test(self, results: Results,
+                          test_suite: str, test_description: str) -> None:
+        # pylint: disable=no-self-use # derived classes may need self
+        """This method runs for each test case that's available and ignored."""
+        results.info('Test case was ignored: {};{}',
+                     test_suite, test_description)
 
     def run(self, results: Results, outcomes: Outcomes) -> None:
         """Check that all available test cases are executed at least once."""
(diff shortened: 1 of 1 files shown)


scripts/mbedtls_framework/outcome_analysis.py
```diff
@@ -192,17 +228,30 @@ class CoverageTask(Task):
     """Analyze test coverage."""
 
     # Test cases whose suite and description are matched by an entry in
-    # IGNORED_TESTS are expected to be never executed.
+    # UNCOVERED_TESTS are expected to be never executed.
+    # Tests matched by IGNORED_TESTS are ignored entierly.
```
ronald-cron-arm:
```suggestion
    # Tests matched by IGNORED_TESTS are ignored entirely.
```

scripts/mbedtls_framework/outcome_analysis.py
```diff
@@ -257,12 +312,17 @@ class DriverVSReference(Task):
     # Configuration name (all.sh component) used as the driver.
     DRIVER = ''
     # Ignored test suites (without the test_suite_ prefix).
-    IGNORED_SUITES = [] #type: typing.List[str]
+    IGNORED_SUITES = [] #type: typing.Sequence[str]
+    # Ignored test cases. Despite the name, these test case are not
+    # completely ignored: they must be skipped by drivers, indicating
+    # a spurious entry.
```
ronald-cron-arm:
```suggestion
    # Ignored test cases. Despite the name, these test cases are not
    # completely ignored: they must be skipped by driver tests. If
    # this is not the case, it indicates a spurious entry.
```

scripts/mbedtls_framework/outcome_analysis.py
```diff
@@ -100,19 +100,61 @@ def execute_reference_driver_tests(results: Results, ref_component: str, driver_
...
+    def __init__(self, description: TestCaseSetDescription) -> None:
+        """Construct a set of test cases from a list of matches for each test suite.
+        """
+        self.matchers = {key: list(value) for key, value in description.items()}
+
+    def extend(self, description: TestCaseSetDescription) -> None:
+        """Add more matchers to this test case set."""
+        for key, entries in description.items():
+            self.matchers.setdefault(key, [])
+            self.matchers[key] += entries
```
ronald-cron-arm, line 129:
```suggestion
            self.matchers[key].extend(entries)
``` ?

File: scripts/mbedtls_framework/outcome_analysis.py [hunk 4]
```diff
@@ -100,19 +100,61 @@ def execute_reference_driver_tests(results: Results, ref_component: str, driver_
...
+    #pylint: disable=too-few-public-methods
+
+    def __init__(self, description: TestCaseSetDescription) -> None:
+        """Construct a set of test cases from a list of matches for each test suite.
+        """
+        self.matchers = {key: list(value) for key, value in description.items()}
+
+    def extend(self, description: TestCaseSetDescription) -> None:
+        """Add more matchers to this test case set."""
+        for key, entries in description.items():
```
ronald-cron-arm, line 127:
```suggestion
        for key, value in description.items():
```
like line 120?

ronald-cron-arm (changes requested):
This looks almost good to me. I only have a few comments. While IGNORED_TESTS is not renamed to UNCOVERED_TESTS in consuming branches, we lose the ability to detect spurious uncovered tests, which seems acceptable.

scripts/mbedtls_framework/outcome_analysis.py
gilles-peskine-arm: What's the difference?

I actually had
```
self.matchers.setdefault(key, []).extend(entries)
```
at one point, because you can't write
```
self.matchers.setdefault(key, []) += extend(entries)
```
because the lhs of the assignment operator has to be a field access, even if it's a compound assignment that isn't actually an assignment. But I felt the two-line version was a little easier to work with. And then I had no reason not to use the operator.

scripts/mbedtls_framework/outcome_analysis.py
gilles-peskine-arm: Here it matters that the value isn't some arbitrary object, but a sequence of entries. Mind you, this also does matter in line 120 as it is now. So I guess rename `value` to `entries` on line 120?

scripts/mbedtls_framework/outcome_analysis.py
ronald-cron-arm: I have a personal preference for `.extend()` thus I suggested it just in case. I am fine to keep `+= entries`.

scripts/mbedtls_framework/outcome_analysis.py
ronald-cron-arm: that sounds good to me

ronald-cron-arm (approved):
LGTM, thanks.

scripts/mbedtls_framework/outcome_analysis.py
```diff
@@ -222,22 +274,28 @@ def run(self, results: Results, outcomes: Outcomes) -> None:
                       suite_case in comp_outcomes.failures
                       for comp_outcomes in outcomes.values())
             (test_suite, test_description) = suite_case.split(';')
-            ignored = self.is_test_case_ignored(test_suite, test_description)
+            ignored = self.ignored_tests.contains(test_suite, test_description)
+            if ignored:
+                self.note_ignored_test(results, test_suite, test_description)
+                continue
```
valeriosetti, line 280: I would have preferred to add the `if ignored` condition to the `if` case below instead of `continue` here, but that's a minor thing.

valeriosetti (approved):
I only have one minor suggestion, but overall LGTM. I will let you decide it it's worth to address it or just merge the PR as is.

PR #296, opened 2026-04-01
Add SSL log validator script

gilles-peskine-arm:
Add a script `scripts/validate_ssl_logs.py` to validate logs from `ssl_client2` and `ssl_server2`. Useful for testing things that we can't test from within a single run of a test programs. Use this to check that when two clients connect to the same server, the connections use different random values.

## PR checklist

- [x] **changelog** not required because: test only
- [ ] **development PR** TODO
- [ ] **4.1 PR** TODO
- [x] **TF-PSA-Crypto PR** not required because: TLS only
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/296
- [x] **3.6 PR** https://github.com/Mbed-TLS/mbedtls/pull/10666
- **tests**  provided


scripts/validate_ssl_logs.py
```diff
@@ -0,0 +1,137 @@
...
+def distinct_server_ephemeral(client_log: ssl_log_parser.Info,
+                              _server_log: ssl_log_parser.Info) -> Optional[str]:
+    """Check that server ephemeral keys as seen from the client are not repeated."""
+    # The current implementation does not handle cases where the client
+    # receives and discards a legitimate resend of the ServerKeyExchange
+    # message in DTLS.
+    if 'DHM: GY' in client_log.dumps:
+        values = client_log.dumps['DHM: GY']
+    else:
+        values = client_log.dumps['server ephemeral public key']
```
bjwtaylor, line 55: If I've understood correctly I believe this will abort if the lookup fails as no matches are present, is this the desired behavior or should it return a validation failure?

scripts/validate_ssl_logs.py
```diff
@@ -0,0 +1,137 @@
...
+    # receives and discards a legitimate resend of the server hello in DTLS.
+    values = client_log.dumps['server hello, random bytes']
+    if len(values) < 2:
+        return 'Fewer than two server_random found'
+    def random_part(hex_data: str) -> str:
+        # In TLS <=1.2, the first 4 bytes (8 hex digits) are the time,
+        # and may differ even if the actually random part is repeated.
+        # The last 8 bytes (16 hex digits) are not random in TLS 1.2 when
+        # the server also supports 1.3 (they are forced to b'DOWNGR\001').
+        return hex_data[8:48]
```
bjwtaylor, line 80: Can this report false duplicates as the last 8 bytes are only non-random in the specific TLS 1.2 downgrade-marker case from RFC 8446?

scripts/validate_ssl_logs.py
```diff
@@ -0,0 +1,137 @@
...
+        outcome = task(client_log, server_log)
+        if outcome is not None:
+            return outcome
+    return None
+
+
+def main() -> int:
+    """Command line entry point."""
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('--list-tasks',
```
bjwtaylor, line 117: Do you need to add action='store_true', action='store_false', or nargs=0, argparse to avoid the error "error: argument --list-tasks: expected one argument" here?

bjwtaylor (dismissed):

bjwtaylor (commented):
As per previous comments

bjwtaylor (commented):
As per previous comments, approved by mistake. Was supposed to be a comment

scripts/validate_ssl_logs.py
gilles-peskine-arm: Good idea, I'll return a more informative and non-fatal validation failure.

scripts/validate_ssl_logs.py
gilles-peskine-arm: The values are supposed to be random, not just unique. So the only way to have a false positive would be if two random 20-byte strings are identical. The probability that this happens is considered low enough in production, so it's good enough in testing as well.

bjwtaylor: OK, sounds reasonable.

bjwtaylor (approved):

bensze01 (approved):
LGTM

PR #297, opened 2026-04-03
Improve fork helper robustness

gilles-peskine-arm:
Resolves most of https://github.com/Mbed-TLS/mbedtls-framework/issues/295, except for the metatests which are added in a consuming branch instead: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/737.

## PR checklist

- [x] **TF-PSA-Crypto development PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/737
- [x] **TF-PSA-Crypto 1.1 PR** not required because: can be updated whenever
- [x] **mbedtls development PR** not required because: does not use fork helper
- [x] **mbedtls 4.1 PR** not required because: does not use fork helper
- [x] **mbedtls 3.6 PR** not required because: can be updated whenever

### DIFF
--- a/tests/include/test/helpers.h
+++ b/tests/include/test/helpers.h
@@ -150,6 +150,13 @@ void mbedtls_test_get_line2(char *line);
 /**
  * \brief           Get a copy of the test result information.
  *
+ * \note This is a shallow copy: in places where the test info structure
+ *       contains a pointer, the pointer is copied. The test framework
+ *       requires these strings to be valid for the duration of the
+ *       test case (including after the test function returns), and does
+ *       not provide any opportunity to deallocate them, so in practice
+ *       they are string literals.
+ *
  * \param[out] out  On output, contains a copy of the current test info.
  */
 void mbedtls_test_info_save(mbedtls_test_info_t *out);
@@ -158,6 +165,13 @@ void mbedtls_test_info_save(mbedtls_test_info_t *out);
  * \brief           Overwrite the test result information.
  *                  This is intended for some unusual scenarios.
  *                  You probably shouldn't use this in a test function.
+ *
+ * \note This is a shallow copy: in places where the test info structure
+ *       contains a pointer, the pointer is copied. The test framework
+ *       requires these strings to be valid for the duration of the
+ *       test case (including after the test function returns), and does
+ *       not provide any opportunity to deallocate them, so in practice
+ *       they are string literals.
  *
  * \param[in] replacement
  *                  The test info to use instead of the current one.
--- a/tests/include/test/helpers.h
+++ b/tests/include/test/helpers.h
@@ -379,6 +379,31 @@ int mbedtls_test_le_s(const char *test, int line_no, const char *filename,
                       long long value1, long long value2);
 
 /**
+ * \brief           Record the current test case as a failure based
+ *                  on a substring search.
+ *
+ *                  This function is usually called via the macro
+ *                  #TEST_STRSTR.
+ *
+ * \param test      Description of the failure or assertion that failed. This
+ *                  MUST be a string literal. This normally has the form
+ *                  "strstr(EXPR1, EXPR2)" where EXPR1 has the value
+ *                  \p haystack and EXPR2 has the value \p needle.
+ * \param line_no   Line number where the failure originated.
+ * \param filename  Filename where the failure originated.
+ * \param haystack  The null-terminated string to look in.
+ *                  Alternatively, this can be a null pointer,
+ *                  which is treated as if it was an empty string.
+ * \param needle    The null-terminated string to look for.
+ *                  Alternatively, this can be a null pointer,
+ *                  which is treated as if it was an empty string.
+ *
+ * \return          \c 1 on success, otherwise \c 0.
+ */
+int mbedtls_test_strstr(const char *test, int line_no, const char *filename,
+                        const char *haystack, const char *needle);
+
+/**
  * \brief          This function decodes the hexadecimal representation of
  *                 data.
  *
--- a/tests/include/test/helpers.h
+++ b/tests/include/test/helpers.h
@@ -345,7 +345,9 @@ mbedtls_threading_mutex_t *mbedtls_test_get_info_mutex(void);
  * \param value1    The first value to compare.
  * \param value2    The second value to compare.
  *
- * \return          \c 1 if the values are equal, otherwise \c 0.
+ * \return          \c 1 if the values are equal.
+ *                  Otherwise, return \c 0 and record the test case as
+ *                  a failure.
  */
 int mbedtls_test_equal(const char *test, int line_no, const char *filename,
                        unsigned long long value1, unsigned long long value2);
@@ -366,7 +368,9 @@ int mbedtls_test_equal(const char *test, int line_no, const char *filename,
  * \param value1    The first value to compare.
  * \param value2    The second value to compare.
  *
- * \return          \c 1 if \p value1 <= \p value2, otherwise \c 0.
+ * \return          \c 1 if \p value1 <= \p value2.
+ *                  Otherwise, return \c 0 and record the test case as
+ *                  a failure.
  */
 int mbedtls_test_le_u(const char *test, int line_no, const char *filename,
                       unsigned long long value1, unsigned long long value2);
@@ -387,7 +391,9 @@ int mbedtls_test_le_u(const char *test, int line_no, const char *filename,
  * \param value1    The first value to compare.
  * \param value2    The second value to compare.
  *
- * \return          \c 1 if \p value1 <= \p value2, otherwise \c 0.
+ * \return          \c 1 if \p value1 <= \p value2.
+ *                  Otherwise, return \c 0 and record the test case as
+ *                  a failure.
  */
 int mbedtls_test_le_s(const char *test, int line_no, const char *filename,
                       long long value1, long long value2);
@@ -412,7 +418,9 @@ int mbedtls_test_le_s(const char *test, int line_no, const char *filename,
  *                  Alternatively, this can be a null pointer,
  *                  which is treated as if it was an empty string.
  *
- * \return          \c 1 on success, otherwise \c 0.
+ * \return          \c 1 if \p needle is a substring of \p haystack.
+ *                  Otherwise, return \c 0 and record the test case as
+ *                  a failure.
  */
 int mbedtls_test_strstr(const char *test, int line_no, const char *filename,
                         const char *haystack, const char *needle);
--- a/tests/include/test/macros.h
+++ b/tests/include/test/macros.h
@@ -121,6 +121,23 @@
         goto exit;                                                  \
     } while (0)
 
+/** Evaluate two string expressions and fail the test case
+ * if the second is not a substring of the first.
+ *
+ * Either value can be null, which is treated as if it was an empy string.
+ *
+ * \param expr1     An string-valued expression to evaluate.
+ * \param expr2     Another string-valued expression to evaluate.
+ */
+#define TEST_STRSTR(expr1, expr2)                                       \
+    do {                                                                \
+        if (!mbedtls_test_strstr("strstr(" #expr1 ", " #expr2 ")",      \
+                                 __LINE__, __FILE__,                    \
+                                 expr1, expr2)) {                       \
+            goto exit;                                                  \
+        }                                                               \
+    } while (0)
+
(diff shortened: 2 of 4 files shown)


tests/src/fork_helpers.c
```diff
@@ -155,90 +157,91 @@ int mbedtls_test_fork_run_child(
...
 
+    /* Create a temporary file where the child can write the test result
+     * and its output. We unlink the file as soon as we've created it,
+     * but we keep it open. This way, both the parent and the child process
+     * can keep using the file until they close it, but the file will
+     * be removed when both processes have closed it.
+     */
+    /* Make up a unique name using the current (parent) process ID and
+     * a stack address inside that process. */
+    char filename[sizeof("mbedtls_test_fork_run_child-%ld-%p.tmp") +
```
bjwtaylor, line 170: I suspect this has probably been consider, but just in case. This is a bit of a regression on the previous implementation as it now requires create/unlink permission, whereas before it could run in read only. Not sure it this will effect anyone though? Maybe if not already considered it might be worth using mkstemp()/tmpfile() as it creates a less predictable filename?

tests/src/fork_helpers.c
```diff
@@ -155,90 +157,91 @@ int mbedtls_test_fork_run_child(
...
-            mbedtls_test_info_overwrite(&reading_on_failure.child_test_info);
-        }
-    } else {
-        /* Weird status, just report it. */
-        TEST_EQUAL(wstatus, 0);
+    TEST_EQUAL(wstatus, CHILD_EXIT_CODE_OK);
+
+    /* The child exited normally. Obtain the test result from the child. */
+    TEST_ASSERT_ERRNO(fseek(file, 0, SEEK_SET) == 0);
+    TEST_ASSERT_ERRNO(fread(&child_test_info, 1, sizeof(child_test_info), file) > 0);
```
bjwtaylor: Would any scenario where less than child_test_info not be a failure here? I think it's minor, however presumably if the file was corrupted or truncated we would want it to fail here?

tests/src/fork_helpers.c
gilles-peskine-arm: We already use files in the current directory for tests that use persistent keys. So yes, it's an additional requirement on the runtime environment compared with the previous fork helper code, but it isn't an additional requirement compared for our crypto testing as a whole.

I deliberately don't use `tmpfile()` or similar to avoid concerns about choosing the right temporary directory, having to worry that it might require unpredictable file names, etc.

tests/src/fork_helpers.c
gilles-peskine-arm: Good point, I meant to read one record of the intended size, but I wrote it the other way round out of habit. I'll fix that.

I didn't bother writing a metatest for a truncated file because it's an unlikely failure that would require us to go out of our way. We don't normally test test code to that extent.

tests/src/fork_helpers.c
bjwtaylor: Ok, sounds reasonable.

bjwtaylor (dismissed):

tests/src/fork_helpers.c
```diff
@@ -67,45 +68,40 @@ static int probably_running_under_valgrind(void)
...
-    }
-
     child_callback(param, buf, size, &length);
-    TEST_LE_U(length, size);
 
-    /* Label called `exit`: this is where TEST_ASSERT() and friends jump to. */
+    if (mbedtls_test_get_result() == MBEDTLS_TEST_RESULT_SUCCESS && length != 0) {
+        /* Write the output. This could fail on a full disk. Remember to
+         * flush (otherwise the output would likely be truncated). */
+        errno = 0;
```
bensze01: Nit: I don't think this is necessary if the implementation of `fwrite`  complies with POSIX.
`TEST_ASSERT_ERRNO` won't read the (stale) value of `errno` if `n == length` since you've switched the order of `1` and `length`

tests/src/fork_helpers.c
```diff
@@ -25,20 +26,20 @@
...
-     * - [1 byte] #mbedtls_test_result_t \c test_result
-     * - Case \c test_result:
-     *     - If \c test_result == #MBEDTLS_TEST_RESULT_SUCCESS:
-     *       the output from the child body function.
-     *     - Otherwise:
-     *       the child failure (or skip) information, a direct write of
-     *       a #mbedtls_test_result_t structure.
+     * The content of the file has the following format:
+     * - mbedtls_test_info_t structure;
+     * - on pass, the ouptut (up to the end of the file).
```
bensze01: Typo
```suggestion
     * - on pass, the output (up to the end of the file).
```

tests/src/fork_helpers.c
bensze01: The output produced by `%p` is implementation defined, so I'm unsure if the size calculation for it is always correct.

How about using `mkstemp()` with the template "mbedtls_test_fork_run_child.tmp-XXXXXX"? We can also add the pid to the template if you'd like - but I'm not sure it's necessary.

tests/src/fork_helpers.c
```diff
@@ -67,45 +68,40 @@ static int probably_running_under_valgrind(void)
...
-        /* There's no way we can report anything other than the exit code.
-         * So we might as well quit without even running the child callback. */
-        goto write_done;
-    }
-
     child_callback(param, buf, size, &length);
-    TEST_LE_U(length, size);
 
-    /* Label called `exit`: this is where TEST_ASSERT() and friends jump to. */
+    if (mbedtls_test_get_result() == MBEDTLS_TEST_RESULT_SUCCESS && length != 0) {
```
bensze01: Nit: ` && length != 0` is unnecessary, since you've switched the positions of the arguments `1` and `length`.

tests/include/test/helpers.h
```diff
@@ -378,6 +378,31 @@ int mbedtls_test_le_u(const char *test, int line_no, const char *filename,
 int mbedtls_test_le_s(const char *test, int line_no, const char *filename,
                       long long value1, long long value2);
 
+/**
+ * \brief           Record the current test case as a failure based
+ *                  on a substring search.
```
bensze01, line 403: Clarify the behaviour of the function
```suggestion
 * \brief           Record the current test case as a failure if
 *                  parameter needle is not a substring of parameter haystack.
```

tests/include/test/helpers.h
```diff
@@ -378,6 +378,31 @@ int mbedtls_test_le_u(const char *test, int line_no, const char *filename,
...
+ * \param line_no   Line number where the failure originated.
+ * \param filename  Filename where the failure originated.
+ * \param haystack  The null-terminated string to look in.
+ *                  Alternatively, this can be a null pointer,
+ *                  which is treated as if it was an empty string.
+ * \param needle    The null-terminated string to look for.
+ *                  Alternatively, this can be a null pointer,
+ *                  which is treated as if it was an empty string.
+ *
+ * \return          \c 1 on success, otherwise \c 0.
```
bensze01: Improve wording
```suggestion
 * \return          \c 0 if the test case was recorded as a failure, otherwise \c 1.
```

tests/include/test/macros.h
```diff
@@ -121,6 +121,23 @@
...
+ * \param expr2     Another string-valued expression to evaluate.
+ */
+#define TEST_STRSTR(expr1, expr2)                                       \
+    do {                                                                \
+        if (!mbedtls_test_strstr("strstr(" #expr1 ", " #expr2 ")",      \
+                                 __LINE__, __FILE__,                    \
+                                 expr1, expr2)) {                       \
+            goto exit;                                                  \
+        }                                                               \
+    } while (0)
```
bensze01: I think this is easier to understand with the parameter names from `mbedtls_test_strstr()`.
```suggestion
 * \param haystack     A string-valued expression to evaluate.
 * \param needle     Another string-valued expression to evaluate.
 */
#define TEST_STRSTR(haystack, needle)                                       \
    do {                                                                \
        if (!mbedtls_test_strstr("strstr(" #haystack ", " #needle ")",      \
                                 __LINE__, __FILE__,                    \
                                 expr1, expr2)) {                       \
            goto exit;                                                  \
        }                                                               \
    } while (0)
```

bensze01 (changes requested):
A few suggestions and typos

tests/src/fork_helpers.c
gilles-peskine-arm: [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fwrite.html#tag_17_199_03) only states that “if a write error occurs, (…) errno shall be set to indicate the error.” It's a classic gotcha that historical implementations don't always set `errno = 0` on success: they may leave it unchanged, or even leave a nonzero result from an internal call that failed in a non-fatal way. Modern implementations usually do set `errno = 0` on success, but I'm old enough to be in the habit of setting it if I care.

tests/src/fork_helpers.c
gilles-peskine-arm: I'd rather not depend on `mkstemp`, which might not exist on embedded platforms where we might want to test `fork()`.

The output of `%p` is unspecified, but I think hex for non-null pointers is really common. But I should check the return value of `snprintf`.

tests/src/fork_helpers.c
bensze01: Sure, I was just pointing out that we don't read the undefined value of `errno` in the success case anyways.

I got the impression that you fixed the issue of `fwrite()` returning `0` if `size == 0` one way (requiring that you overwrite `errno`), then switched the `size` and `count` parameters around so now `n == length` is always true and we no longer need to reset `errno` (like we don't anywhere else in this file).

tests/include/test/helpers.h
gilles-peskine-arm: I don't see how this is more informative, since the readers doesn't know the parameters yet when they read the brief description.

gilles-peskine-arm:
There are a lot of CI failures, probably because this branch is too old. I'm going to rebase on top of the head of `main`. The last reviewed commit is f61edfd639390fea15756361d83466078ee000e8 → dcf3f2b248b8bd07321e52c8c6505994958ac17a

tests/src/fork_helpers.c
```diff
@@ -155,90 +156,90 @@ int mbedtls_test_fork_run_child(
...
+        mbedtls_test_info_overwrite(&child_test_info);
+        goto exit;
     }
 
+    /* The child succeeded. Read the output. */
+    size_t len = fread(child_output, 1, child_output_size, file);
+    /* Error out on read error */
+    TEST_ASSERT_ERRNO(!ferror(file));
+    /* Error out if the child wrote more than child_output_size bytes */
+    int c = getc(file);
```
bjwtaylor, line 230: Does this also need to check the ferror?

tests/src/fork_helpers.c
```diff
@@ -155,90 +156,90 @@ int mbedtls_test_fork_run_child(
...
+     * and its output. We unlink the file as soon as we've created it,
+     * but we keep it open. This way, both the parent and the child process
+     * can keep using the file until they close it, but the file will
+     * be removed when both processes have closed it.
+     */
+    /* Make up a unique name using the current (parent) process ID and
+     * a stack address inside that process. */
+    char filename[sizeof("mbedtls_test_fork_run_child-%ld-%p.tmp") +
+                  3 * sizeof(long) + 2 * sizeof(void *)];
+    TEST_LE_S(1, mbedtls_snprintf(filename, sizeof(filename),
```
bjwtaylor: Does this account for truncation similar to this https://github.com/Mbed-TLS/mbedtls-framework/pull/297#discussion_r3766902665

tests/src/fork_helpers.c
```diff
@@ -155,90 +156,90 @@ int mbedtls_test_fork_run_child(
...
+                                  "mbedtls_test_fork_run_child-%ld-%p.tmp",
+                                  (long) getpid(), (void *) filename));
+    file = fopen(filename, "w+");
+    TEST_ASSERT_ERRNO(file != NULL);
+    mbedtls_test_fork_child_fd = fileno(file);
+    unlink(filename);
```
bjwtaylor: Should the return be checked?

gilles-peskine-arm:
The CI is failing on TF-PSA-Crypto development. I need to rebase this branch again to get a meaningful CI run. The last commit that Ben reviewed is fd4cef7ad21033282b79476f7db526cacb5b5bf2 → 49e3acd53b4700ef50db65f5fc67733c1f130ac7

PR #305, opened 2026-05-18
Allow macros that shadow mldsa-native functions

bensze01:
## Description

`check_names.py` normally requires uppercase macro names. Make an exception for helper macros that shadow the functions names used by mldsa-native.

## PR checklist

- [ ] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/305
- [ ] **TF-PSA-Crypto development PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/786
- [ ] **TF-PSA-Crypto 1.1 PR**  not required because: new feature work
- [ ] **mbedtls development PR**  not required because: crypto-only feature work
- [ ] **mbedtls 4.1 PR** not required because: crypto-only feature work
- [ ] **mbedtls 3.6 PR** not required because: crypto-only feature work

### DIFF
--- a/scripts/check_names.py
+++ b/scripts/check_names.py
@@ -61,7 +61,12 @@ PUBLIC_MACRO_PATTERN = re.compile(r"^(MBEDTLS|PSA|TF_PSA_CRYPTO)_[0-9A-Z_]*[0-9A
 # implementation), must not end with an underscore (for legibility),
 # and must not consist of a single letter (because we've seen embedded
 # platforms that claim single-uppercase-letter names for themselves).
-INTERNAL_MACRO_PATTERN = re.compile("^[A-Z][0-9A-Z_]*[A-Z0-9]$")
+#
+# We make an exception for the helper macros that shadow the functions
+# declared by mldsa-native.
+INTERNAL_MACRO_PATTERN = re.compile(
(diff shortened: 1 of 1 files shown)


scripts/check_names.py
```diff
@@ -61,7 +61,12 @@
 # implementation), must not end with an underscore (for legibility),
 # and must not consist of a single letter (because we've seen embedded
 # platforms that claim single-uppercase-letter names for themselves).
-INTERNAL_MACRO_PATTERN = re.compile("^[A-Z][0-9A-Z_]*[A-Z0-9]$")
+#
+# We make an exception for the helper macros that shadow the functions
```
gilles-peskine-arm, line 65: I'd prefer to make the exception more specific and do it via to `name_pattern_exception`. Is there a reason why that wouldn't work?

gilles-peskine-arm (commented):
Ok in principle, but can we make the exception more specific?

bjwtaylor (commented):
Looks good except for the issue Gilles pointed out, the commit message also has a grammatical error “functions names” should be “function names”

bensze01:
This patch is no longer needed by the TF-PSA-Crypto PR.

PR #307, opened 2026-06-11
scripts: Add dispatch include path and new generation location

asmellby:
## Description

PSA crypto driver dispatch files are generated into `dispatch/` instead of `core/` in newer TF-PSA-Crypto versions. API headers that rely on driver and dispatch implementation details are also moved from `include/` to `dispatch/include/`.

Update the framework scripts to work with both the old and new layouts.

## PR checklist

- [x] **TF-PSA-Crypto development PR** provided Mbed-TLS/TF-PSA-Crypto#809
- [x] **TF-PSA-Crypto 1.1 PR** not required because: not a bugfix, repository structure change is not entirely backwards compatible
- [x] **mbedtls development PR** provided Mbed-TLS/mbedtls#10780
- [x] **mbedtls 4.1 PR** not required because: does not affect mbedtls
- [x] **mbedtls 3.6 PR** not required because: does not affect mbedtls

### DIFF
--- a/scripts/make_generated_files.py
+++ b/scripts/make_generated_files.py
@@ -22,11 +22,16 @@ COMMON_GENERATION_SCRIPTS = [
 ]
 
 if build_tree.looks_like_tf_psa_crypto_root("."):
+    new_dispatch_layout = (Path.cwd() / "dispatch" / "CMakeLists.txt").exists()
+    if build_tree.tf_psa_crypto_version() >= (1,1,0) and new_dispatch_layout:
+        dispatch_prefix = Path("dispatch")
+    else:
+        dispatch_prefix = Path("core")
     TF_PSA_CRYPTO_GENERATION_SCRIPTS = [
         GenerationScript(
             Path("scripts/generate_driver_wrappers.py"),
-            [Path("core/psa_crypto_driver_wrappers.h"),
-             Path("core/psa_crypto_driver_wrappers_no_static.c")],
+            [dispatch_prefix / "psa_crypto_driver_wrappers.h",
+            dispatch_prefix / "psa_crypto_driver_wrappers_no_static.c"],
             "", None
         ),
         GenerationScript(
--- a/scripts/make_generated_files.py
+++ b/scripts/make_generated_files.py
@@ -22,16 +22,15 @@ COMMON_GENERATION_SCRIPTS = [
 ]
 
 if build_tree.looks_like_tf_psa_crypto_root("."):
-    new_dispatch_layout = (Path.cwd() / "dispatch" / "CMakeLists.txt").exists()
-    if build_tree.tf_psa_crypto_version() >= (1,1,0) and new_dispatch_layout:
-        dispatch_prefix = Path("dispatch")
+    if (Path.cwd() / "dispatch" / "CMakeLists.txt").exists():
+        DISPATCH_PREFIX = Path("dispatch")
     else:
-        dispatch_prefix = Path("core")
+        DISPATCH_PREFIX = Path("core")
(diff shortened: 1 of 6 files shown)


gilles-peskine-arm:
(Priority note: this fits under our “PSA crypto driver standardization” work, for which we have allocated a little bandwidth this quarter. Thanks Aksel for contributing it!)

scripts/make_generated_files.py
```diff
@@ -22,11 +22,16 @@
 ]
 
 if build_tree.looks_like_tf_psa_crypto_root("."):
+    new_dispatch_layout = (Path.cwd() / "dispatch" / "CMakeLists.txt").exists()
+    if build_tree.tf_psa_crypto_version() >= (1,1,0) and new_dispatch_layout:
```
gilles-peskine-arm: What's the point of checking the version? I'd prefer to avoid relying on any version number.

scripts/test_psa_constant_names.py
```diff
@@ -171,6 +171,7 @@ def main():
     else:
         parser.add_argument('--include', '-I',
                             action='append', default=['tf-psa-crypto/include',
+                                                      'tf-psa-crypto/dispatch/include',
```
gilles-peskine-arm, line 174: There's still a reference to `core/psa_crypto_driver_wrappers.h` in `scripts/check_names.py`, and even a reference to `library/psa_crypto_driver_wrappers.h` which only exists in the 3.6 branch. It was introduced in 634e0d25cb and I think it's now obsolete (especially since `check_names.py` passes just fine with the generated files in `dispatch`), but I'm not completely sure. If it's obsolete, we might as well remove it now, although it would be strictly speaking out of scope.

gilles-peskine-arm (changes requested):
I don't think `tf_psa_crypto_version()` is necessary. But if it is, it needs to pass the static checks. Otherwise looks good to me.

scripts/make_generated_files.py
asmellby: I did it to limit when the new layout was considered to releases that are "supposed to" have the new layout. But unconditionally allowing the new layout does mean we avoid the version check. Updated to remove the version check.

scripts/test_psa_constant_names.py
asmellby: Good catch. Updated to include the new dispatch include path (which was missing) and generalizing the excluded autogen files. I agree that it looks like the script passes even if `psa_crypto_driver_wrappers.h` isn't excluded, so if you want me to, I can remove the exclusion instead. I don't know if the lack of "bad" symbols in the generated file is a long-term commitment or just how things are right now.

gilles-peskine-arm: The problem isn't with symbols that violate our naming conventions, but about symbols that look like a typo because `check_names` finds their use but not their definition or vice versa.

Since 634e0d25cb9e75c54c72391be28da4c2f7a1def0, we've both fixed a few bugs in `check_names.py` and changed the structure of the driver wrapper code. Today `check_names.py` passes on 3.6 if I remove all references to `psa_crypto_driver_wrappers.h` (all of them in exclude lists) (69fa1da16dede0eeab60451a08549546bfbe32f9). So please go ahead and do that.

gilles-peskine-arm (commented):
LGTM, but let's not add to the complexity in `check_names`, please remove the `psa_crypto_driver_wrappers.h` exclusion.

scripts/test_psa_constant_names.py
asmellby: Done.

gilles-peskine-arm (approved):
LGTM at 957b8a1b7492054a18eee5f20e3f7f626be06503

PR #313, opened 2026-07-09
generate_test_cert_macros.py: Removed absolute path from the rendered output

minosgalanakis:
## Description

As part of the post release activities, fixing minor inconveniences of the tooling required to build a release.

This pr adjust the generate_test_cert_macros.py logic to use relative paths BEFORE rendering the templates for the auto-generated files but continue using absolute paths internally so it can be called from different locations in the tree.

Needs to be tested locally with `./framework/scripts/generate_test_cert_macros.py` and inspection of the header files.

## PR checklist

- [ ] **TF-PSA-Crypto development PR** not required because: script contained here
- [ ] **TF-PSA-Crypto 1.1 PR**  not required because: script contained here
- [ ] **mbedtls development PR** not required because: script contained here
- [ ] **mbedtls 4.1 PR** not required because: script contained here
- [ ] **mbedtls 3.6 PR** not required because: script contained here

### DIFF
--- a/scripts/generate_test_cert_macros.py
+++ b/scripts/generate_test_cert_macros.py
@@ -15,9 +15,9 @@ import argparse
 import jinja2
 from mbedtls_framework.build_tree import guess_project_root
 
-TESTS_DIR = os.path.join(guess_project_root(), 'tests')
-FRAMEWORK_DIR = os.path.join(guess_project_root(), 'framework')
-DATA_FILES_PATH = os.path.join(FRAMEWORK_DIR, 'data_files')
+PROJECT_ROOT = guess_project_root()
+TESTS_DIR = os.path.join(PROJECT_ROOT, 'tests')
+DATA_FILES_PATH = 'framework/data_files'
 
 INPUT_ARGS = [
     ("string", "TEST_CA_CRT_EC_PEM", DATA_FILES_PATH + "/test-ca2.crt"),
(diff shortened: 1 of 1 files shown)


bjwtaylor (approved):

bjwtaylor (approved):
The changes look good but there is a ci failure that will need to be resolved.

ronald-cron-arm:
Why do you mention make_generated_files.py here where it is about  scripts/generate_test_cert_macros.py?

ronald-cron-arm:
Otherwise could you clean-up the PR description? Remove the "Notes for the submitter", update the PR checklist.

minosgalanakis:
> Why do you mention make_generated_files.py here where it is about scripts/generate_test_cert_macros.py?

Because generate_test_cert_macros is called by make_generated_files, but yes you are right it could be more specific. 

Also the CI failure is the typical intermitten ssl one, restarted

ronald-cron-arm:
> > Why do you mention make_generated_files.py here where it is about scripts/generate_test_cert_macros.py?
> 
> Because generate_test_cert_macros is called by make_generated_files, but yes you are right it could be more specific.

Please update the title and PR description accordingly.

ronald-cron-arm (changes requested):
See https://github.com/Mbed-TLS/mbedtls-framework/pull/313#issuecomment-4933987792

gilles-peskine-arm (changes requested):
There is already a fix for that in https://github.com/Mbed-TLS/mbedtls-framework/pull/231. Let's not spend more time making a second one.

PR #316, opened 2026-08-11
Prerequisites for testing Blake2 algorithm in TF-PSA-Crypto

valeriosetti:
## Description

Add support for Blake2 hash algorithms.
Prerequisite for https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867.

## PR checklist

- [ ] **TF-PSA-Crypto development PR** provided: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867
- [ ] **TF-PSA-Crypto 1.1 PR** not required because: please see https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867
- [ ] **mbedtls development PR** not required because: please see https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867
- [ ] **mbedtls 4.1 PR** not required because: please see https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867
- [ ] **mbedtls 3.6 PR** not required because: please see https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867

### DIFF
--- a/scripts/mbedtls_framework/crypto_data_tests.py
+++ b/scripts/mbedtls_framework/crypto_data_tests.py
@@ -57,6 +57,8 @@ class HashPSALowLevel:
         'PSA_ALG_SHAKE256_256': None, #lambda data: hashlib.shake_256(data).hexdigest(32),
         'PSA_ALG_SHAKE256_512': None, #lambda data: hashlib.shake_256(data).hexdigest(64),
         'PSA_ALG_ASCON_HASH256': None, #lambda data: ascon.ascon_hash(data),
+        'PSA_ALG_BLAKE2S_HASH256': lambda data: hashlib.blake2s(data).hexdigest(),
+        'PSA_ALG_BLAKE2B_HASH512': lambda data: hashlib.blake2b(data).hexdigest(),
     } #type: Dict[str, Optional[Callable[[bytes], str]]]
 
     @staticmethod
--- a/scripts/mbedtls_framework/crypto_data_tests.py
+++ b/scripts/mbedtls_framework/crypto_data_tests.py
@@ -59,7 +59,7 @@ class HashPSALowLevel:
         'PSA_ALG_ASCON_HASH256': None, #lambda data: ascon.ascon_hash(data),
         'PSA_ALG_BLAKE2S_HASH256': lambda data: hashlib.blake2s(data).hexdigest(),
         'PSA_ALG_BLAKE2B_HASH512': lambda data: hashlib.blake2b(data).hexdigest(),
-        'PSA_ALG_SM3': None, #lambda data: hashlib.new('sm3').hexdigest(),
+        'PSA_ALG_SM3': None, #lambda data: hashlib.new('sm3', data).hexdigest(),
     } #type: Dict[str, Optional[Callable[[bytes], str]]]
 
     @staticmethod
--- a/scripts/mbedtls_framework/crypto_data_tests.py
+++ b/scripts/mbedtls_framework/crypto_data_tests.py
@@ -59,6 +59,7 @@ class HashPSALowLevel:
         'PSA_ALG_ASCON_HASH256': None, #lambda data: ascon.ascon_hash(data),
         'PSA_ALG_BLAKE2S_HASH256': lambda data: hashlib.blake2s(data).hexdigest(),
         'PSA_ALG_BLAKE2B_HASH512': lambda data: hashlib.blake2b(data).hexdigest(),
+        'PSA_ALG_SM3': None, #lambda data: hashlib.new('sm3').hexdigest(),
     } #type: Dict[str, Optional[Callable[[bytes], str]]]
 
     @staticmethod
--- a/scripts/mbedtls_framework/crypto_knowledge.py
+++ b/scripts/mbedtls_framework/crypto_knowledge.py
@@ -353,6 +353,8 @@ class Algorithm:
     CATEGORY_FROM_HEAD = {
         'AES_MMO_ZIGBEE': AlgorithmCategory.HASH,
         'ASCON_HASH': AlgorithmCategory.HASH,
+        'BLAKE2B': AlgorithmCategory.HASH,
+        'BLAKE2S': AlgorithmCategory.HASH,
         'SHA': AlgorithmCategory.HASH,
         'SHAKE256_512': AlgorithmCategory.HASH,
         'MD': AlgorithmCategory.HASH,
@@ -360,6 +362,7 @@ class Algorithm:
         'SM3': AlgorithmCategory.HASH,
         'ANY_HASH': AlgorithmCategory.HASH,
         'HMAC': AlgorithmCategory.MAC,
+        'BLAKE2_MAC': AlgorithmCategory.MAC,
         'STREAM_CIPHER': AlgorithmCategory.CIPHER,
         'ASCON_AEAD': AlgorithmCategory.AEAD,
         'CHACHA20_POLY1305': AlgorithmCategory.AEAD,
--- a/scripts/mbedtls_framework/crypto_knowledge.py
+++ b/scripts/mbedtls_framework/crypto_knowledge.py
@@ -485,13 +485,13 @@ class Algorithm:
             return False
         return kdf_alg in self.KEY_DERIVATIONS_INCOMPATIBLE_WITH_AGREEMENT
 
-    HASH_ALGS_WITHOUT_OID = [
+    HASH_ALGS_WITHOUT_OID = frozenset([
         'PSA_ALG_BLAKE2S_HASH256',
         'PSA_ALG_BLAKE2B_HASH512',
         'PSA_ALG_AES_MMO_ZIGBEE',
         'PSA_ALG_ASCON_HASH256',
         'PSA_ALG_SHAKE256_512',
-    ]
+    ])
     def is_valid_rsa_alg_with_hash(self) -> bool:
         """Whether the specified combinaton of RSA_PKCS1V15_SIGN() and hash is supported.
         Rationale: there are hash algorithms for which OID is not standardized (or not yet),
--- a/scripts/mbedtls_framework/crypto_knowledge.py
+++ b/scripts/mbedtls_framework/crypto_knowledge.py
@@ -241,7 +241,13 @@ class KeyType:
            alg.head == 'STREAM_CIPHER':
             return True
         if self.head == 'RSA' and alg.head.startswith('RSA_'):
-            return alg.is_valid_rsa_alg_with_hash()
+            # There are hash algorithms for which OID is not standardized (or not yet),
+            # so RSA PKCS#1 v1.5 signature is not supported. For the full list of unsupported
+            # hash algorithms please refer to the official PSA API documentation:
+            # https://arm-software.github.io/psa-api/crypto/1.5/api/ops/signature.html#c.PSA_ALG_RSA_PKCS1V15_SIGN
+            if alg.head == 'RSA_PKCS1V15_SIGN':
+                return alg.get_inner_algorithm().has_hash_standardized_oid()
+            return True
         if alg.category == AlgorithmCategory.KEY_AGREEMENT and \
            self.is_public():
             # The PSA API does not use public key objects in key agreement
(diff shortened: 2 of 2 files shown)


scripts/generate_psa_tests.py
```diff
@@ -480,6 +483,14 @@ def exercise_key_with_algorithm(
         m = cls.BRAINPOOL_RE.match(key_type.string)
         if m and alg.string != 'PSA_ALG_ECDSA_ANY':
             return False
+        # RSA PKCS#1v1.5 signatures embed an ASN.1 OID identifying the hash
+        # algorithm, but there is no such OID being officially standardized for
+        # for BLAKE2. This is also reported on the PSA Crypto API specification:
+        # https://arm-software.github.io/psa-api/crypto/1.5/api/ops/hash.html#c.PSA_ALG_BLAKE2S_HASH256
```
bjwtaylor, line 489: I believe this is now incorrect as these are specified here https://www.rfc-editor.org/rfc/rfc7693.html#section-4 Let me know what you think though?

gilles-peskine-arm: This is correct with respect to the PSA specification.

I'll go and check if that OID is actually used in RSA PKCS#1v1.5 signatures. (It's sometimes a bit of a mess.) If so we'll update the PSA specification.

valeriosetti: > I'll go and check if that OID is actually used in RSA PKCS#1v1.5 signatures. (It's sometimes a bit of a mess.) If so we'll update the PSA specification.

@gilles-peskine-arm do you have any update on this?

gilles-peskine-arm: OpenSSL doesn't support RSA PKCS#1v1.5 with BLAKE2. If there isn't already an implementation using the OIDs from RFC 7693 for RSA PKCS#1v1.5, I'd rather not be the first one that does. We aren't planning to list these OIDs in the PSA specification unless either we find an existing implementation, or there is a business case for RSA PKCS#1v1.5 with BLAKE2 using the PSA API.

scripts/generate_psa_tests.py
```diff
@@ -443,6 +443,9 @@ def __init__(self, info: psa_information.Information, version: int, forward: boo
 
     RSA_OAEP_RE = re.compile(r'PSA_ALG_RSA_OAEP\((.*)\)\Z')
     BRAINPOOL_RE = re.compile(r'PSA_KEY_TYPE_\w+\(PSA_ECC_FAMILY_BRAINPOOL_\w+\)\Z')
+    RSA_PKCS1V15_SIGN_BLAKE2_RE = \
```
gilles-peskine-arm: BLAKE2s and BLAKE2B aren't the only algorithms for which there's no RSA PKCS#1v1.5 signature. Please make this check more structured, based on a new method in `Algorithm` class of the `crypto_knowledge` module, similar to `is_valid_key_agreement_with_derivation()`.

scripts/generate_psa_tests.py
```diff
@@ -480,6 +483,14 @@ def exercise_key_with_algorithm(
         m = cls.BRAINPOOL_RE.match(key_type.string)
         if m and alg.string != 'PSA_ALG_ECDSA_ANY':
             return False
+        # RSA PKCS#1v1.5 signatures embed an ASN.1 OID identifying the hash
+        # algorithm, but there is no such OID being officially standardized for
+        # for BLAKE2. This is also reported on the PSA Crypto API specification:
+        # https://arm-software.github.io/psa-api/crypto/1.5/api/ops/hash.html#c.PSA_ALG_BLAKE2S_HASH256
+        if key_type.string in ('PSA_KEY_TYPE_RSA_KEY_PAIR',
+                               'PSA_KEY_TYPE_RSA_PUBLIC_KEY') and \
```
gilles-peskine-arm: Why check the key type? The algorithm itself is invalid.

gilles-peskine-arm (changes requested):
LGTM, except please code the RSA exception in Python, not in perl-with-python-syntax

scripts/generate_psa_tests.py
gilles-peskine-arm: And while you're at it, please list the [other exceptions](https://arm-software.github.io/psa-api/crypto/1.5/api/ops/signature.html#c.PSA_ALG_RSA_PKCS1V15_SIGN), even if we don't implement them yet: `PSA_ALG_AES_MMO_ZIGBEE`, `PSA_ALG_ASCON_HASH256 PSA_ALG_SHAKE256_512`.

valeriosetti: > And while you're at it, please list the [other exceptions](https://arm-software.github.io/psa-api/crypto/1.5/api/ops/signature.html#c.PSA_ALG_RSA_PKCS1V15_SIGN), even if we don't implement them yet: `PSA_ALG_AES_MMO_ZIGBEE`, `PSA_ALG_ASCON_HASH256 PSA_ALG_SHAKE256_512`.

Sure. Shouldn't Blake2 be listed there as well? Should I open a PR toward the documentation to fix that?

gilles-peskine-arm: Oh, yes, please. The specification of each blake variant says no PKCS1v1.5, but the PKCS1v1.5 list of exceptions is missing blake.

scripts/generate_psa_tests.py
valeriosetti: Since this was a standalone commit and the [previous comment](https://github.com/Mbed-TLS/mbedtls-framework/pull/316#discussion_r3936293758) suggested a better way to handle this, I removed the commit and added a new one with the (hopefully) correct implementation.

scripts/generate_psa_tests.py
valeriosetti: FYI: this is the PR fixing the documentation https://github.com/GlobalPlatform/psa-api/pull/382. I cannot add you as reviewer though :/

scripts/mbedtls_framework/crypto_data_tests.py
```diff
@@ -57,6 +57,9 @@ def __init__(self, info: psa_information.Information) -> None:
         'PSA_ALG_SHAKE256_256': None, #lambda data: hashlib.shake_256(data).hexdigest(32),
         'PSA_ALG_SHAKE256_512': None, #lambda data: hashlib.shake_256(data).hexdigest(64),
         'PSA_ALG_ASCON_HASH256': None, #lambda data: ascon.ascon_hash(data),
+        'PSA_ALG_BLAKE2S_HASH256': lambda data: hashlib.blake2s(data).hexdigest(),
+        'PSA_ALG_BLAKE2B_HASH512': lambda data: hashlib.blake2b(data).hexdigest(),
+        'PSA_ALG_SM3': None, #lambda data: hashlib.new('sm3').hexdigest(),
```
gilles-peskine-arm: @ronald-cron-arm Do all the python implementations on our CI (including Windows) have SM3?

(Ronald noted that while SM3 is not in `hashlib` in Python itself, it may be provided by the OpenSSL backend.)

scripts/mbedtls_framework/crypto_knowledge.py
```diff
@@ -482,6 +485,29 @@ def is_invalid_key_agreement_with_derivation(self) -> bool:
             return False
         return kdf_alg in self.KEY_DERIVATIONS_INCOMPATIBLE_WITH_AGREEMENT
 
+    HASH_ALGS_WITHOUT_OID = [
```
gilles-peskine-arm: Minor: since this is just to test membership:

```suggestion
    HASH_ALGS_WITHOUT_OID = frozenset([
```

scripts/mbedtls_framework/crypto_knowledge.py
```diff
@@ -241,7 +241,7 @@ def can_do(self, alg: 'Algorithm') -> bool:
            alg.head == 'STREAM_CIPHER':
             return True
         if self.head == 'RSA' and alg.head.startswith('RSA_'):
-            return True
+            return alg.is_valid_rsa_alg_with_hash()
```
gilles-peskine-arm: The function is also true for algorithms that involve RSA but no hash, or no RSA, so the name is misleading. I think the auxiliary function should only be called when `alg.head == 'r'RSA_PKCS1V15_SIGN'`.

gilles-peskine-arm (changes requested):
Good shape, I just want some minor changes to the execution

scripts/mbedtls_framework/crypto_data_tests.py
ronald-cron-arm: I do not know if all the Python implementations on our CI have SM3. Let me check.

gilles-peskine-arm: On Ubuntu 18.04 and above, `/usr/bin/python3` has sm3. But the images built by pyenv on my machine don't have it, so I suspect that python in the Ubuntu 16.04 dockerfile doesn't have it either. It's probably possible to have it with pyenv, but we'd need to rebuild it with the right options, and maybe with an additional build-time prerequisite.

ronald-cron-arm: Ubuntu 16.04 comes with OpenSSL 1.0.2g which does not support SM3 it seems. I am okay with the None anyway.

ronald-cron-arm:
```suggestion
        'PSA_ALG_SM3': None, #lambda data: hashlib.new('sm3', data).hexdigest(),
```

gilles-peskine-arm (dismissed):
LGTM at 252169b8566f1261789e96da8b1bc147938d5355

gilles-peskine-arm:
This will need to be rebased on top of `main` to pass the CI, now that #321 has been merged.

scripts/mbedtls_framework/crypto_knowledge.py
```diff
@@ -482,6 +491,27 @@ def is_invalid_key_agreement_with_derivation(self) -> bool:
...
+        'PSA_ALG_ASCON_HASH256',
+        'PSA_ALG_SHAKE256_512',
+    ])
+    def has_hash_standardized_oid(self) -> bool:
+        """Whether the hash algorithm has a standardized OID."""
+        if self.expression in self.HASH_ALGS_WITHOUT_OID:
+            return False
+        return True
+
+    def get_inner_algorithm(self) -> "Algorithm":
```
ronald-cron-arm, line 507:
```suggestion
    def get_inner_algorithm(self) -> Optional["Algorithm"]:
```
because of "return None" line 511?

ronald-cron-arm (commented):
just one thing I am not sure about

scripts/mbedtls_framework/crypto_knowledge.py
valeriosetti: Oh, right. I don't think using `Optional` is the solution here given how I'm using the returned value from `get_inner_algorithm` (i.e. `.get_inner_algorithm().has_hash_standardized_oid()`). I think the best option is to return an exception here. I mean, there is no reason to complicate the management, if you call `get_inner_algorithm` you should be reasonably sure that the format of the algorithm is correct; if it's not then returning an exception is the easiest solution.
Wdyt?

ronald-cron-arm: I am okay with an exception

valeriosetti: @ronald-cron-arm I just pushed https://github.com/Mbed-TLS/mbedtls-framework/pull/316/commits/fcec9520fa2fb411a59a44339781ba0c65825a6c to give you an idea of what I explained above. Let me know what you think about it ;)

gilles-peskine-arm (approved):
LGTM at fcec9520fa2fb411a59a44339781ba0c65825a6c

valeriosetti:
This last force push is to rebase on `main`. I forgot to do that before. Sorry

scripts/mbedtls_framework/crypto_knowledge.py
ronald-cron-arm: all good to me, thanks.

ronald-cron-arm (approved):
LGTM @https://github.com/Mbed-TLS/mbedtls-framework/commit/edeb57f200ef92c01dcdf77fa90616ed038ebcf5

gilles-peskine-arm (approved):
LGTM

PR #320, opened 2026-09-04
Adapt some test scripts for SM3 support

ronald-cron-arm:
## Description
Needed for https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/895

## PR checklist
- [x] **TF-PSA-Crypto development PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/895
- [x] **TF-PSA-Crypto 1.1 PR** not required because: new feature development only
- [x] **mbedtls development PR** not required because: crypto only 
- [x] **mbedtls 4.1 PR** not required because: crypto only
- [x] **mbedtls 3.6 PR** not required because: crypto only

### DIFF
--- a/scripts/generate_psa_tests.py
+++ b/scripts/generate_psa_tests.py
@@ -443,6 +443,7 @@ class StorageFormat:
 
     RSA_OAEP_RE = re.compile(r'PSA_ALG_RSA_OAEP\((.*)\)\Z')
     BRAINPOOL_RE = re.compile(r'PSA_KEY_TYPE_\w+\(PSA_ECC_FAMILY_BRAINPOOL_\w+\)\Z')
+    SM3_COMBINATION_RE = re.compile(r'PSA_ALG_\w+\(PSA_ALG_SM3\)\Z')
     @classmethod
     def exercise_key_with_algorithm(
             cls,
@@ -463,6 +464,10 @@ class StorageFormat:
         # Raw data keys have no useful exercise anyway so there is no
         # loss of test coverage.
         if key_type.string == 'PSA_KEY_TYPE_RAW_DATA':
+            return False
+        # SM3 support is limited to hash operations. Do not exercise key
+        # algorithms that use SM3 as a component.
+        if cls.SM3_COMBINATION_RE.match(alg.string):
             return False
         # OAEP requires room for two hashes plus wrapping
         m = cls.RSA_OAEP_RE.match(alg.string)
--- a/scripts/mbedtls_framework/crypto_data_tests.py
+++ b/scripts/mbedtls_framework/crypto_data_tests.py
@@ -16,6 +16,15 @@ from . import psa_test_case
 from . import test_case
 
 
+def _hashlib_calculator(name: str) -> Optional[Callable[[bytes], str]]:
+    """Return a hashlib-based calculator if the algorithm is available."""
+    try:
+        hashlib.new(name)
+    except ValueError:
+        return None
+    return lambda data: hashlib.new(name, data).hexdigest()
+
+
 class HashPSALowLevel:
     """Generate test cases for the PSA low-level hash interface."""
 
@@ -56,6 +65,7 @@ class HashPSALowLevel:
         'PSA_ALG_SHAKE128_256': None, #lambda data: hashlib.shake_128(data).hexdigest(32),
         'PSA_ALG_SHAKE256_256': None, #lambda data: hashlib.shake_256(data).hexdigest(32),
         'PSA_ALG_SHAKE256_512': None, #lambda data: hashlib.shake_256(data).hexdigest(64),
+        'PSA_ALG_SM3': _hashlib_calculator('sm3'),
         'PSA_ALG_ASCON_HASH256': None, #lambda data: ascon.ascon_hash(data),
     } #type: Dict[str, Optional[Callable[[bytes], str]]]
 


scripts/mbedtls_framework/crypto_data_tests.py
```diff
@@ -16,6 +16,15 @@
 from . import test_case
 
 
+def _hashlib_calculator(name: str) -> Optional[Callable[[bytes], str]]:
+    """Return a hashlib-based calculator if the algorithm is available."""
+    try:
+        hashlib.new(name)
```
gilles-peskine-arm, line 22: Having output that depends on the version of Python could be a problem. At one point, I'd like to clean up our Python testing, and one of the things I would test is that different versions of Python produce the same generated file.

scripts/mbedtls_framework/crypto_data_tests.py
```diff
@@ -56,6 +65,7 @@ def __init__(self, info: psa_information.Information) -> None:
         'PSA_ALG_SHAKE128_256': None, #lambda data: hashlib.shake_128(data).hexdigest(32),
         'PSA_ALG_SHAKE256_256': None, #lambda data: hashlib.shake_256(data).hexdigest(32),
         'PSA_ALG_SHAKE256_512': None, #lambda data: hashlib.shake_256(data).hexdigest(64),
+        'PSA_ALG_SM3': _hashlib_calculator('sm3'),
```
gilles-peskine-arm, line 68: Why bother to look for it? SM3 is not [in the current development version of Python](https://docs.python.org/3.16/library/hashlib.html).

I'd like to move to a model where only maintainers generate this test data, and all algorithms are mandatory. We now have this model for `generate_mldsa_tests.py`. We should do the same for hashes, but that's at least a size-s task of itself.

scripts/generate_psa_tests.py
```diff
@@ -464,6 +465,10 @@ def exercise_key_with_algorithm(
         # loss of test coverage.
         if key_type.string == 'PSA_KEY_TYPE_RAW_DATA':
             return False
+        # SM3 support is limited to hash operations. Do not exercise key
+        # algorithms that use SM3 as a component.
```
gilles-peskine-arm, line 469: This is a temporary limitation in code that is not in the framework. At some point we're going to want to lift it, but then we'll need a migration path for the time period where the same framework commit is consumed both by TF-PSA-Crypto branches that have the limitation and branches that haven't.

One solution for this transition is to first add support for compound operations in TF-PSA-Crypto, then wait some time (at least one release, I would say) to flush the branches we care about, and finally lift the limitation here. But this takes time.

A better solution is to split the script into a small runner in the consuming project, and code in the framework that takes some arguments that determine what to generate. This is what I've done with [`generate_mldsa_tests.py`](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/tf-psa-crypto-1.2.0/scripts/maintainer/generate_mldsa_tests.py) and [`mldsa_test_generator.py`](https://github.com/Mbed-TLS/mbedtls-framework/blob/mbedtls-4.2.0_tf-psa-crypto-1.2.0/util/mbedtls_maintainer/mldsa_test_generator.py). But migrating to this architecture would be a size-s task of itself.

Another possible solution would be to have the script detect the availability of the algorithms, like it detects the availability of the basic building blocks. I don't know if there's a plausible way to do that. The test case generator parses `test_suite_psa_crypto_metadata.data` to find combinations of algorithms, in addition to having rules to build combinations from constructors. That code is a bit of a mess, and I don't know offhand if there's a reasonable way to extract meaningful information.

Yet another solution would be to generate all the tests as usual, and make it the responsibility of TF-PSA-Crypto to skip some tests at runtime. It's a bit of a hack, but I think it's fine for a temporary solution. We'd have to list exceptions in `analyze_outcomes.py`, and we'd remember to stop skipping these tests thanks to the issue linked from the exception.

Finally, a clean solution would be to not add code to prevent algorithm combinations in the first place. This has my preference. Why should we go out of our way to block things that will naturally work? The only extra work is to define an HMAC block size. You don't have to make a big pull request with everything: you can add tests in further pull requests.

gilles-peskine-arm (changes requested):
There are two ticking bombs. The python version sensitivity is not likely to explode until way after SM3 lands, but it also seems unnecessary. The restriction on generated tests will explode while SM3 lands.

gilles-peskine-arm:
https://github.com/Mbed-TLS/mbedtls-framework/pull/316 is closer to what's needed. I think it would be easier for you both if one of you made a single PR that lists both Blake and SM3.

scripts/mbedtls_framework/crypto_data_tests.py
ronald-cron-arm: Here it is just that if the `hashlib.new('sm3')` is supported we generate automatically some SM3 tests, otherwise we do not. If we can generate them why not doing it?

scripts/mbedtls_framework/crypto_data_tests.py
ronald-cron-arm: With `hashlib.new('sm3')` we have support for SM3 through OpenSSL on out-of-the-box Ubuntu 24.04 if I understand correctly. "Why bother" : we have the tests for free.

scripts/generate_psa_tests.py
ronald-cron-arm: What I had in mind was to avoid having support for SM3 itself and for all the algorithms that can use SM3 as their hash in the same PR. The idea was simply to spread the implementation and review effort across multiple PRs. This seems to create more problems than it solves, so I am going to do everything in the first PR to avoid the difficulties described above.

ronald-cron-arm:
> #316 is closer to what's needed. I think it would be easier for you both if one of you made a single PR that lists both Blake and SM3.

fine by me

scripts/mbedtls_framework/crypto_data_tests.py
gilles-peskine-arm: I don't want the generated files to depend on the platform. That's guaranteed to bite us sooner or later.

ronald-cron-arm:
We will go with #316.

